PR #2, opened 2021-02-11
Tracking lesson progress

xswoodsx:
### Context
Teachers need to be able to track their progress for individual lessons, to be able to do this we need to link up early career teachers and course lessons. 

- Adding migrations that create CourseLessonProgress table and add a uniqueness constraint so as not to generate duplicate records of a teachers progress against the same lesson.
- Updated Schema
- Adding model tests for CourseLessonProgresses, along with a Factory
- Additional association tests for EarlyCareerTeacherProfile model

### Guidance to review
I've only included two course lesson progress states for the minute. These are likely to be added too later on. 

### Testing
As mentioned above.

### DIFF
--- a/db/migrate/20210209172653_create_course_lesson_progresses.rb
+++ b/db/migrate/20210209172653_create_course_lesson_progresses.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateCourseLessonProgresses < ActiveRecord::Migration[6.1]
+  def change
+    create_table :course_lesson_progresses, id: :uuid do |t|
+      t.string :progress, default: "not_started"
+      t.references :early_career_teacher_profile, null: false, foreign_key: true, index: { name: :idx_course_lesson_progresses_on_ect_profile_id }
+      t.references :course_lesson, null: false, foreign_key: true, index: { name: :idx_course_lesson_progresses_on_course_lesson_id }
+      t.timestamps
+    end
+  end
+end
--- a/spec/models/course_lesson_progress_spec.rb
+++ b/spec/models/course_lesson_progress_spec.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe CourseLessonProgress, type: :model do
+  it "is created with the default progress value" do
+    course_lesson = CourseLessonProgress.new
+    expect(course_lesson[:progress]).to eql("not_started")
+  end
+
+  describe "associations" do
+    it { is_expected.to belong_to(:course_lesson) }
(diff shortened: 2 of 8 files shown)


spec/models/course_lesson_progress_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+          early_career_teacher_profile: teacher,
+          course_lesson: lesson_one,
+        )
+        expect(lesson_one_progress_duplicate).not_to be_valid
+      end
+
+      it "creates a second course lesson progress record" do
+        create(:course_lesson_progress, early_career_teacher_profile: teacher, course_lesson: lesson_two)
+        expect(teacher.course_lesson_progresses.count).to eql(2)
+      end
```
kicferk1, line 59: It seems to duplicate `"adds a second lesson record to a teachers progresses"` test. Let's axe it :)

db/migrate/20210209172653_create_course_lesson_progresses.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateCourseLessonProgresses < ActiveRecord::Migration[6.1]
+  def change
+    create_table :course_lesson_progresses, id: :uuid do |t|
+      t.string :progress, default: "not_started"
```
amazimbe, line 6: If this only has a fixed set of values then probably we should use an enum rather than a string field. https://api.rubyonrails.org/v5.2.4.4/classes/ActiveRecord/Enum.html

kicferk1: I prefer to store enums as strings over integers, but that solution looks like it would work: https://stackoverflow.com/a/32939934

ethanmills: +1 for string enums

ethanmills: Preferably stored as an enum in the db

kicferk1: Something like https://til.hashrocket.com/posts/b492fb26ec-rails-enums-and-postgresql-enums ?

I think this is more effort than it's worth. We will be changing this many times over (I think), and I don't want each change to deal with the postgres type. Also I don't like to rely on database-specific implementations, especially when they go over the head of the ORM we are using.

Plus, I would be shocked if Rails understands postgres types and outputs them into schema file. And I quite like being able to see the schema and know what is in the db.

Also, with an enum, trying to assign an incorrect value will raise a nice exception, so I think it's pretty safe and we will catch it early enough.

kicferk1 (approved):

amazimbe (approved):

PR #3, opened 2021-02-12
Add breadcrumbs

kicferk1:
### Context
I want CIP navigation to be easier

Add breadcrumbs using govuk-components to our cip pages. 'home' should go to dashboard when user is logged in, and to cip page otherwise.

Add list of modules to ect dashboard.

### Guidance to review
Commit by commit, maybe check out the review app. `early-career-teacher@example.com` should log you in to the dashboard. 

### Testing
We will probably need to add some cypress tests for it later.

### DIFF
--- a/app/views/core_induction_programme/lessons/show.html.erb
+++ b/app/views/core_induction_programme/lessons/show.html.erb
@@ -1,7 +1,4 @@
-<%= govuk_breadcrumbs breadcrumbs: [
-  ["Home", @current_user ? dashboard_path : cip_path],
-  [@course_lesson.course_module.course_year.title, cip_year_path(@course_lesson.course_module.course_year)],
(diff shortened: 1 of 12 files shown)


app/views/core_induction_programme/lessons/show.html.erb
```diff
@@ -1,3 +1,7 @@
+<%= govuk_breadcrumbs breadcrumbs: [
+  ["Home", @current_user ? dashboard_path : cip_path],
+  [@course_lesson.course_module.course_year.title, cip_year_path(@course_lesson.course_module.course_year)],
+  [@course_lesson.course_module.title, cip_year_module_path(@course_lesson.course_module.course_year, @course_lesson.course_module)]] %>
```
callumacrae, line 4: This is feeling a bit duplicatey, is there any way to clean this up easily?

kicferk1: Eh, I guess we could have a helper for it. Will do :)

callumacrae (approved):

PR #5, opened 2021-02-15
Add mentor profile

xswoodsx:
### Context 
There will be mentors for ECF that will need to access the service. This adds the initial creation of these users.

- Migrations to add MentorProfile and linking ECTProfiles to a mentor
- Updated Schema
- MentorProfile model tests
- Updated ECTProfile model tests
- Creating a Mentor in dummy seed data, with a ECT being assigned a Mentor
- Mentor email added to test users to log in automatically.

### Testing
As above

### DIFF
--- a/app/models/early_career_teacher_profile.rb
+++ b/app/models/early_career_teacher_profile.rb
@@ -5,7 +5,6 @@ class EarlyCareerTeacherProfile < ApplicationRecord
   belongs_to :core_induction_programme, optional: true
   belongs_to :cohort, optional: true
   belongs_to :mentor_profile, optional: true
-
   has_one  :mentor, through: :mentor_profile, source: :user
 
   has_many :course_lesson_progresses
--- a/app/models/early_career_teacher_profile.rb
+++ b/app/models/early_career_teacher_profile.rb
@@ -4,6 +4,9 @@ class EarlyCareerTeacherProfile < ApplicationRecord
   belongs_to :user
   belongs_to :core_induction_programme, optional: true
   belongs_to :cohort, optional: true
+  belongs_to :mentor_profile, optional: true
+
+  has_one  :mentor, through: :mentor_profile, source: :user
 
   has_many :course_lesson_progresses
   has_many :course_lessons, through: :course_lesson_progresses
--- a/app/views/dashboard/show.html.erb
+++ b/app/views/dashboard/show.html.erb
@@ -2,15 +2,23 @@
   <div class="govuk-grid-column-full">
     <h1 class="govuk-heading-l">User dashboard</h1>
     <p class="govuk-body"> You are logged in</p>
-
     <% if @current_user.early_career_teacher? %>
       <h2 class="govuk-heading-l">Your Core Induction Programme</h2>
-      <% if @current_user.course_years.each do |course_year| %>
+      <% @current_user.course_years.each do |course_year| %>
         <p><%= govuk_link_to course_year.title, cip_year_url(course_year) %></p>
-      <% end.empty? %>
+      <% end %>
+      <% if @current_user.course_years.empty? %>
         <p>You have not been enrolled in Core Induction Programme yet!</p>
       <% end %>
     <% end %>
-
+    <%if @current_user.mentor? %>
+      <h2 class="govuk-heading-l">Your list of mentees</h2>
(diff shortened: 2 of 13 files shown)


app/models/early_career_teacher_profile.rb
```diff
@@ -4,6 +4,9 @@ class EarlyCareerTeacherProfile < ApplicationRecord
   belongs_to :user
   belongs_to :core_induction_programme, optional: true
   belongs_to :cohort, optional: true
+  belongs_to :mentor_profile, optional: true
+
+  has_one  :mentor, through: :mentor_profile, source: :user
```
kicferk1, line 9: I would group these two lines together - logical connection trumps semantics and syntax.

spec/factories/mentor_profiles.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :mentor_profile do
+    user
+  end
+end
```
kicferk1, line 7: Can we also add a trait to the user factory for the case of them being a mentor?

kicferk1 (commented):
Can we also display list of mentees on the dashboard page? That will be proof enough that our seeding works :)

app/views/dashboard/show.html.erb
```diff
@@ -2,15 +2,23 @@
...
     <p class="govuk-body"> You are logged in</p>
-
     <% if @current_user.early_career_teacher? %>
       <h2 class="govuk-heading-l">Your Core Induction Programme</h2>
-      <% if @current_user.course_years.each do |course_year| %>
+      <% @current_user.course_years.each do |course_year| %>
         <p><%= govuk_link_to course_year.title, cip_year_url(course_year) %></p>
-      <% end.empty? %>
+      <% end %>
+      <% if @current_user.course_years.empty? %>
```
kicferk1, line 10: Pity we are changing that, I quite liked the solution.

But vscode code formatting is too much of a pain in the ass to tell you that not all ends are created equal.

kicferk1 (approved):

kicferk1:
Looks like there is a confliiiiict. Happy to help with it if it is terrible.

app/views/dashboard/show.html.erb
```diff
@@ -2,15 +2,23 @@
...
         <p>You have not been enrolled in Core Induction Programme yet!</p>
       <% end %>
     <% end %>
-
+    <%if @current_user.mentor? %>
+      <h2 class="govuk-heading-l">Your list of mentees</h2>
+      <% @current_user.mentor_profile.early_career_teachers.each do |early_career_teacher| %>
+        <p><%= early_career_teacher.full_name %></p>
+      <% end %>
+      <% if @current_user.mentor_profile.early_career_teachers.empty? %>
```
ethanmills, line 19: @amazimbe am I right in thinking this could be written as `end.empty?` ? Should it be?

spec/factories/users.rb
```diff
@@ -19,5 +19,9 @@
     trait :early_career_teacher do
       early_career_teacher_profile
     end
+
+    trait :mentor do
+      mentor_profile
```
ethanmills, line 24: It turns out this creates a different user, then steals their mentor profile 😱 
Instead, I've been using `mentor_profile { build(:mentor_profile) }`. Slightly less pretty, but it doesn't mess up the cypress tests so much

app/views/dashboard/show.html.erb
xswoodsx: Not sure why but VSCode indentation goes out of wack when we do end.empty

spec/factories/users.rb
kicferk1: what xD WHAT??? WHAAAAAAAT?????

Fine, and I imagine we should do it for all of our traits?

ethanmills: Yup 😢

app/views/dashboard/show.html.erb
kicferk1: It could be.

Steven's auto formatter doesn't believe it should be though. And delving into depths of VSCode plugins is the last thing I want to do. 

When he gets a ruby mine license we will probably rewrite it.

ethanmills: I think line 7 would need it's `if` back. Not important though, was mostly just curious 🙂

kicferk1 (approved):

PR #6, opened 2021-02-18
Add Cypress

callumacrae:
It's pretty much the same as the setup from the other repo, just with slightly different tests.

### DIFF
--- a/spec/cypress/integration/aaa-meta.js
+++ b/spec/cypress/integration/aaa-meta.js
@@ -1,0 +1,73 @@
+/**
+ * !!!!!
+ *
+ * This file name starts with aaa so that it runs before all the other tests
+ * which rely on the functionality being tested in this file.
+ *
+ * Please don't rename it!
+ */
+describe("Meta test helper tests", () => {
+  it("should have login and logout helper commands", () => {
+    cy.login();
+    cy.get("h1").should("contain", "User dashboard");
+
+    cy.logout();
+    cy.get("#success-message").should("contain", "Signed out successfully.");
+  });
+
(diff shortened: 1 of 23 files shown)


spec/cypress/integration/aaa-meta.js
```diff
@@ -0,0 +1,73 @@
...
+    cy.get('[href*="/modules/"]').contains("Test Course module").click();
+
+    cy.get("@courseLesson").then(([course]) => {
+      cy.url().should("contain", `/modules/${course.course_module_id}`);
+    });
+
+    cy.get('[href*="/lessons/"]').contains("Test Course lesson").click();
+
+    cy.get("@courseLesson").then(([course]) => {
+      cy.url().should("contain", `/lessons/${course.id}`);
```
kicferk1, line 40: I think there is something funky going on with these urls.

```
/modules/${course.course_module_id}
```
should be 
```
/modules/${course.id}
```
cause a course does not have `course_module_id` on it (lessons do).

And 
```
/lessons/${course.id}
```
should really be more like 
```
/lessons/${lesson.id}
```
Might be just variable names that are weird, although the first one looks a bit more fishy.

callumacrae: There is something funky going on with the variable names - for some reason I shortened `course lesson` to `course`, not `lesson` (or maybe it used to be the course). Will fix!

(Also, this was approved code in the other repo 😆)

kicferk1: There is no way it can work in other repo - it was merged after we removed course lessons :D

kicferk1 (approved):

PR #8, opened 2021-02-18
Lesson content model

kicferk1:
### Context
We want to be able to split the content of a lesson into several parts. That's because the content is really quite long.

1. Add `CourseLessonPart` model
2. Move `CourseLesson` content inside that model
3. Update cip seed files and scripts to match new content type
4. Lesson page redirects to the first content page
5. Add show and edit pages for lesson parts.

### Guidance to review
As usual cip seed files are long and horrible - I made sure they make sense. The main change is moving content from lessons to lesson parts.

I haven't added a way to add more than one lesson part on frontend yet, and so I haven't added any content to the lesson part page that would relate to multiple parts - links to previous / next, list of parts, etc.

### Testing
1. Unit tests
2. No tests
3. No tests
4. Unit testing lesson page requests
5. Unit testing show and edit lesson part page requests
Might break cypress tests I guess. I should probably pull from develop.


spec/factories/course_lesson.rb
```diff
@@ -4,5 +4,11 @@
   factory :course_lesson do
     title { "Test Course lesson" }
     course_module { FactoryBot.create(:course_module) }
+
+    trait :with_lesson_part do
```
callumacrae, line 8: If the tests were failing, does that mean that this is required? If so, should it be a trait or should it just happen anyway?

(sorry for commenting on a PR you haven't requested a review on yet lol)

kicferk1: Tests were failing because in case a lesson doesn't have a part, I don't show any content. And the tests, well, they expect content 😂

app/models/core_induction_programme_exporter.rb
```diff
@@ -8,22 +8,46 @@ def run
...
     SeedDump.dump(
-      CourseYear.all,
+      course_modules,
       file: "db/seeds/cip_seed_dump.rb",
       exclude: %i[created_at updated_at],
       import: true,
       append: true,
     )
+
+    course_lessons = []
```
amazimbe, line 33: We could use reduce /  inject. Probably more rubyscue

kicferk1: Ruby has two methods for reducing? :O 

I went with `map` + `flatten`. I think it's easier to understand than reduce.

amazimbe (approved):

app/models/core_induction_programme_exporter.rb
amazimbe: Yes, the usual ruby thing about coder happiness. Apparently, inject doesn't feel natural to advocates of reduce. I think one is just an alias to the other though.

app/models/course_lesson_part.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class CourseLessonPart < ApplicationRecord
+  belongs_to :course_lesson
+
+  has_one :next_lesson_part, class_name: "CourseLessonPart", foreign_key: :previous_lesson_part_id
+  belongs_to :previous_lesson_part, class_name: "CourseLessonPart", inverse_of: :next_lesson_part, optional: true
+
+  validates :title, presence: { message: "Enter a title" }
```
seb-siem-softwire, line 9: Do we want to restrict the number of characters? For example id db you are storing it as a string and if you reach more than 256 characters...it will just do nothing as far as I remember. If we don't want to restrict - changing the field to text will allow stpring more characters. Also we could validate on putting just alphanumeric characters

seb-siem-softwire: around 7 years ago I was working on a project that had string in the db schema that led to bipolar behaviour and tracking it down - lets say allowed me to fully embrace abusive vocabulary

app/views/core_induction_programme/lesson_parts/show.html.erb
```diff
@@ -0,0 +1,28 @@
+<%= govuk_breadcrumbs breadcrumbs: course_lesson_breadcrumbs(@current_user, @course_lesson_part.course_lesson) %>
+<div class="govuk-grid-row">
```
seb-siem-softwire, line 2: Just a suggestion - adding couple of extra empty lines will speed up reading the file - cognitively we store information in chunks and then form relationships between those chunks

app/views/core_induction_programme/lessons/show.html.erb
```diff
@@ -2,23 +2,9 @@
...
-              [OpenStruct.new(id: :complete, name: "I feel confident that I've read and understood this session")],
-              :id,
-              :name,
-              legend: {text: "That's the end of this session"}) %>
-        <%= f.govuk_submit "Complete session" %>
-      <% end %>
+      <%= link_to "Edit lesson", edit_cip_year_module_lesson_path(id: @course_lesson.id), class: "govuk-button" %>
     <% end %>
+    <h1 class="govuk-heading-xl"> <%= @course_lesson.title %></h1>
+    <p>This lesson has no content!</p>
```
seb-siem-softwire, line 8: Well that's a bummer.... hahahaha

db/migrate/20210217160425_add_lesson_content_part.rb
```diff
@@ -0,0 +1,15 @@
...
+class AddLessonContentPart < ActiveRecord::Migration[6.1]
+  def change
+    create_table :course_lesson_parts do |t|
+      t.timestamps
+
+      t.column :title, :string, null: false
+      t.column :content, :text, null: false, limit: 100_000
+
+      t.references :previous_lesson_part, null: true, foreign_key: { to_table: :course_lesson_parts }, type: :uuid
+      t.references :course_lesson, null: false, foreign_key: true, type: :uuid
```
seb-siem-softwire, line 12: Do we want to add any indexes? Open question - just a thought

spec/policies/course_lesson_part_policy_spec.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe CourseLessonPartPolicy, type: :policy do
+  subject { described_class.new(user, course_lesson_part) }
```
seb-siem-softwire, line 6: you can also add named subject for example subject(:foo) - just a thought

rupollbabu (commented):
Numbers are not in sequence, Requested @kicferk1 fix it

app/views/core_induction_programme/lessons/show.html.erb
kicferk1: Hopefully this page never gets shown 😂 
But hey, who knows, right? :P

db/migrate/20210217160425_add_lesson_content_part.rb
kicferk1: I think foreign key gives us index for free.

My schema has
```
t.index ["course_lesson_id"], name: "index_course_lesson_parts_on_course_lesson_id"
t.index ["previous_lesson_part_id"], name: "index_course_lesson_parts_on_previous_lesson_part_id"
```
which are pretty much the things I want to index on.

spec/policies/course_lesson_part_policy_spec.rb
kicferk1: Good to know :)

db/migrate/20210217160425_add_lesson_content_part.rb
seb-siem-softwire: 👍

spec/policies/course_lesson_part_policy_spec.rb
seb-siem-softwire: 👍

seb-siem-softwire (approved):

kicferk1:
> Numbers are not in sequence, Requested @kicferk1 fix it

They should be fixed now :)

PR #9, opened 2021-02-19
Add more cypress tests

kicferk1:
### Context
I wanted to spend a chunk of my day on a calm tech-debt ticket :)
Also, I hope this will make @rupollbabu happier.

Add cypress test showing what admin can do and what ECT interaction with CIP is.

### Guidance to review
If they pass they pass, eh?
I wasn't really sure where best to put the methods for visiting year / module / lesson. I made them into cypress commands, but if you prefer them to live somewhere else let me know.

### Testing
Eh, this whole PR is just tests for the sake of tests, I don't think there is much testing we can do on them :)

### DIFF
--- a/spec/cypress/integration/admin-cip-stories.js
+++ b/spec/cypress/integration/admin-cip-stories.js
@@ -4,7 +4,6 @@ describe("Admin user interaction with Core Induction Programme", () => {
   });
 
   it("should show a download export button", () => {
-    cy.login("admin");
     cy.visit("/core-induction-programme");
     cy.get("a.govuk-button").contains("Download export").should("exist");
   });
--- a/spec/cypress/integration/admin-cip-stories.js
+++ b/spec/cypress/integration/admin-cip-stories.js
@@ -1,0 +1,95 @@
+describe("Admin user interaction with Core Induction Programme", () => {
+  beforeEach(() => {
+    cy.login("admin");
+  });
+
+  it("should show a download export button", () => {
+    cy.login("admin");
+    cy.visit("/core-induction-programme");
+    cy.get("a.govuk-button").contains("Download export").should("exist");
+  });
+
+  it("should allow to edit year title", () => {
+    cy.appFactories([["create", "course_year"]]);
+
+    cy.visitYear();
+    cy.get("h1").should("contain", "Test Course year");
+    cy.get("a.govuk-button").contains("Edit year content").should("exist");
+    cy.get("a.govuk-button").contains("Edit year content").click();
+
+    cy.get("h1").should("contain", "Content change preview");
+    cy.get("input[name='title']").type("New title");
+    cy.get("input").contains("See preview").click();
+
+    cy.get("h1").should("contain", "Content change preview");
+    cy.visitYear();
+    cy.get("h1").should("contain", "Test Course year");
+
+    cy.get("a.govuk-button").contains("Edit year content").click();
+    cy.get("input[name='title']").type("New title");
+    cy.get("input").contains("Save changes").click();
+
+    cy.get("h1").should("contain", "New title");
+  });
+
+  it("should allow to edit module title", () => {
+    cy.appFactories([["create", "course_module"]]);
+
+    cy.visitModule();
+    cy.get("h1").should("contain", "Test Course module");
+    cy.get("a.govuk-button").contains("Edit module content").should("exist");
+    cy.get("a.govuk-button").contains("Edit module content").click();
+
+    cy.get("h1").should("contain", "Content change preview");
+    cy.get("input[name='title']").type("New title");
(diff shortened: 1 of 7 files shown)


spec/cypress/integration/admin-cip-stories.js
```diff
@@ -0,0 +1,94 @@
...
+  it("should show a download export button", () => {
+    cy.visit("/core-induction-programme");
+    cy.get("a.govuk-button").contains("Download export").should("exist");
+  });
+
+  it("should allow to edit year title", () => {
+    cy.appFactories([["create", "course_year"]]);
+
+    cy.visitYear();
+    cy.get("h1").should("contain", "Test Course year");
```
callumacrae, line 15: In my opinion this (and all the assertions like it) would be better inside the `visit*` commands

spec/cypress/integration/admin-cip-stories.js
```diff
@@ -0,0 +1,94 @@
...
+
+    cy.get("h1").should("contain", "New title");
+  });
+
+  it("should allow to edit module title", () => {
+    cy.appFactories([["create", "course_module"]]);
+
+    cy.visitModule();
+    cy.get("h1").should("contain", "Test Course module");
+    cy.get("a.govuk-button").contains("Edit module content").should("exist");
```
callumacrae, line 39: This isn't needed - `.contains()` implies `.should('exist')` unless `.should('not.exist')` is explicitly specified, so the line below covers this

spec/cypress/integration/admin-cip-stories.js
```diff
@@ -0,0 +1,94 @@
...
+    cy.appFactories([["create", "course_year"]]);
+
+    cy.visitYear();
+    cy.get("h1").should("contain", "Test Course year");
+    cy.get("a.govuk-button").contains("Edit year content").should("exist");
+    cy.get("a.govuk-button").contains("Edit year content").click();
+
+    cy.get("h1").should("contain", "Content change preview");
+    cy.get("input[name='title']").type("New title");
+    cy.get("input").contains("See preview").click();
```
callumacrae, line 21: I'm uncomfortable using element tag names as selectors in feature-focused tests. This could change to a button and the user wouldn't care, but the test would start failing.

Cypress have some recommendations on their best practice page around selectors: https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements

callumacrae:
nice! 👍

one more extremely minor point - i'd be tempted to put the file names the other way around (admin-cip-stories -> cip-stories-admin) so that when there are more tests they're alphabetically ordered next to each other

spec/cypress/integration/admin-cip-stories.js
kicferk1: Eh, maybe you are right.

On the other hand, we are using govuk form builder to create submits, so I imagine it won't change any time soon.

I switched to just using text, and it seems to work.

spec/cypress/integration/cip-stories-early-career-teacher.js
```diff
@@ -0,0 +1,55 @@
...
+
+    cy.visitLesson();
+    cy.visitModule();
+    cy.contains("strong.govuk-tag", "in progress");
+
+    cy.visitLesson();
+    cy.get('[type="checkbox"]').check();
+    cy.get("form").submit();
+    cy.contains("strong.govuk-tag", "complete");
+  });
```
callumacrae, line 54: sorry for not thinking about this in my first review, but the way the helpers you've created work and are being used is a bit of an anti-pattern. e.g., with this test you are doing:

1. visit year
2. [assertion]
3. visit year
4. visit module
5. visit lesson
6. visit year
7. visit module
8. [assertion]
9. visit year
10. visit module
11. visit lesson
12. [assertions]

Also, if you create more than one lesson/module, the helper will fail, as it won't know which link to click on.

I can't immediately think of an actually good solution, but one thought I've had is that `visitLesson()` could behave the same way as `login()` does - it could create a lesson, and then immediately navigate to the lesson it just created using the URL. Then any navigating between levels (e.g. from lesson to module) could be done without a helper, by clicking on the back link.

rupollbabu (approved):
**End-End UI Test covered for following roles**
Admin and early-career-teacher ---roles
**Features**
Edit year title
Edit year content
Edit module title
Edit module content
Edit lesson title
Edit lesson content

early-career-teacher ---roles
Status---not started , in progress, completed.

spec/cypress/integration/cip-stories-early-career-teacher.js
kicferk1: I pushed an update - feel free to have a look :)

callumacrae (approved):

PR #10, opened 2021-02-22
Determine module progress

xswoodsx:
### Context
We need to add the course module progresses for users on to the dashboard and course years page. Module progresses are not stored in the db, instead we look at course lessons progress to determine if a module is 'not_started', "in_progress" or "completed". Any lesson in a module with a status as "in_progress", should show the module as "in_progress. Any module with all lessons as "complete" or "not_started" should show the module progress as the same status.

- Setting a users module progress in course year. 
- Reflecting these progresses in the CourseYear show and Dashboard show
- Retrieving Lessons and a users LessonProgresses has been split out into a separate helper.

### Guidance to review
- The CourseYearHelper and CourseYear model have been refactored in `520379`

### Testing
See commit `3d0adc8`


kicferk1:
Oops, looks like there are conflicts. Happy to pair on them if they are horrible.

app/helpers/course_year_helper.rb
```diff
@@ -0,0 +1,14 @@
...
+  def get_user_lessons_and_progresses(ect_profile)
+    course_lessons = CourseLesson.where(course_module: course_modules)
+
+    user_progresses = CourseLessonProgress.where(early_career_teacher_profile: ect_profile, course_lesson: course_lessons).includes(:course_lesson)
+
+    course_lessons.map do |lesson|
+      lesson.progress = user_progresses.find { |progress| progress.course_lesson == lesson }&.progress || "not_started"
+      lesson
+    end
+  end
```
kicferk1, line 13: I think this method should be commonised with the one in `course_module` - the only difference is what set of course lessons they take, and that could be an argument.

app/models/course_module.rb
```diff
@@ -27,4 +29,16 @@ def lessons_with_progress(user)
...
+    lessons = lessons_with_progress(user)
+    if lessons.all? { |lesson| lesson.progress == "not_started" }
+      return "not_started"
+    end
+    if lessons.all? { |lesson| lesson.progress == "complete" }
+      return "complete"
+    end
+
+    "in_progress"
+  end
```
kicferk1, line 43: Are we using this?

app/views/core_induction_programme/years/show.html.erb
```diff
@@ -12,9 +12,26 @@
...
+            </span>
+            <%= render ProgressLabelComponent.new progress: course_module.progress %>
+          </li>
+        <% end.empty? %>
+        <p>No course modules were found!</p>
+      <% end %>
+    </ul>
+  </li>
+</ol>
+</div>
```
kicferk1, line 36: Indentation woes due to VSCode and `<% end.empty? %>` line :/

app/models/course_module.rb
xswoodsx: doesn't look like we are. i'll remove it and give the app a run

app/views/core_induction_programme/years/show.html.erb
xswoodsx: VSWoes. Usual case of it saving and moving it back.

app/helpers/course_year_helper.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module CourseYearHelper
+  def get_user_lessons_and_progresses(ect_profile, course_lessons = nil)
+    course_lessons ||= CourseLesson.where(course_module: course_modules)
```
kicferk1, line 5: That is a funny line. I imagine `course_modules` will not be recognised inside `CourseModule` :)
I would get rid of it, and pass course lessons to this method in both cases.

app/models/course_module.rb
```diff
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 
 class CourseModule < ApplicationRecord
+  attr_accessor :progress
   include OrderHelper
+  include CourseYearHelper
```
kicferk1, line 6: I would put the accessor after the `includes`.

Aaand `CourseYearHelper` is not something I expect inside `course_module`. Maybe rename it to `CourseLessonProgressHelper`?

app/models/course_year.rb
```diff
@@ -17,4 +18,32 @@ def course_modules_in_order
...
+    modules_in_order = course_modules_in_order
+    ect_profile = user&.early_career_teacher_profile
+    return modules_in_order unless ect_profile
+
+    set_user_course_module_progresses(get_user_lessons_and_progresses(ect_profile))
+  end
+
+private
+
+  def set_user_course_module_progresses(course_lessons)
```
kicferk1, line 32: `compute_user_course_module_progress`?

app/models/course_year.rb
```diff
@@ -17,4 +18,32 @@ def course_modules_in_order
...
+    course_modules.map do |course_module|
```
kicferk1, line 33: Probably want `modules_in_order` here - and I would expect them to be passed to the method to avoid re-computing them.

app/views/core_induction_programme/years/show.html.erb
```diff
@@ -6,15 +6,25 @@
...
+              <%= render ProgressLabelComponent.new progress: course_module.progress %>
+            </li>
+          <% end.empty? %>
+          <p>No course modules were found!</p>
+        <% end %>
+      </ul>
+    </li>
+  </ol>
+</div>
 </div>
```
kicferk1, line 30: Still looks a bit off :/

xswoodsx: i've tried sorting this. Removing the .empty? and have separate blocks but there's still something it doesn't like. Saving keeps formatting it to as above. Will keep trying (I'm chasing for a RubyMine license)

spec/models/course_year_spec.rb
```diff
@@ -27,21 +27,52 @@
...
       course_module_three.update!(previous_module: course_module_four)
 
-      expected_lessons_with_order = [course_module_one, course_module_two, course_module_four, course_module_three]
-      course_year.course_modules_in_order.zip(expected_lessons_with_order).each do |actual, expected|
+      expected_lessons_with_order = [@course_module_one, course_module_two, course_module_four, course_module_three]
+      @course_year.course_modules_in_order.zip(expected_lessons_with_order).each do |actual, expected|
         expect(actual).to eq(expected)
       end
     end
+    it "returns a progress of not_started when all lessons in that module are not_started" do
```
kicferk1, line 60: Spaces between tests please :)

spec/models/course_year_spec.rb
```diff
@@ -27,21 +27,52 @@
...
+      module_progress = @course_year.modules_with_progress(@teacher).first
+      expect(module_progress.progress).to eql("not_started")
+    end
+    it "returns a progress of in_progress when one lesson in that module is in_progress" do
+      @course_lesson_one_progress.update!(progress: "in_progress")
```
kicferk1, line 65: `@ course_lesson_one_progress.in_progress!` will do the same. As Alfred put it, power of enum :O

spec/models/course_year_spec.rb
```diff
@@ -27,21 +27,52 @@
...
+
+      module_progress = @course_year.modules_with_progress(@teacher).first
+      expect(module_progress.progress).to eql("in_progress")
+    end
+    it "returns a progress of complete when all lessons in that module are complete" do
+      @course_lesson_one_progress.update!(progress: "complete")
+      @course_lesson_two_progress.update!(progress: "complete")
```
kicferk1, line 72: Ditto, enums are great.

kicferk1 (commented):
Looks pretty good, just stylistic comments :)

kicferk1 (approved):

PR #13, opened 2021-02-24
Split lesson parts

kicferk1:
### Context
Lessons are long. We want to split them.
Here is the UI to split them.

1. Add a button and a page for splitting a lesson part.
2. Add endpoints for splitting a lesson part.
3. Add controller methods and policies for splitting a lesson part.

### Guidance to review
I haven't added lesson part navigation yet. That's next PR.

I opted to use more space than govuk pages use on the split page. That's because I think it is really convenient to see the old and new lesson part side-by-side. And for that, 960px that gov pages believe in, is too narrow. I intend to show it to Mark A (who will likely be the person who interacts with it the most) and see if he likes it.

image

### Testing
1. Not testing the presence of the button.
2. Tested with request tests (rspec).
3. Tested with request tests and unit tests (rspec).

### DIFF
--- a/app/views/core_induction_programme/lesson_parts/show_split.html.erb
+++ b/app/views/core_induction_programme/lesson_parts/show_split.html.erb
@@ -1,5 +1,19 @@
 <%= govuk_breadcrumbs breadcrumbs: course_lesson_breadcrumbs(@current_user, @course_lesson_part.course_lesson) %>
-<h1 class="govuk-heading-l">Content change preview</h1>
+<h1 class="govuk-heading-l">Split lesson part</h1>
+<ul>
+  <li>
+    The part on the left hand side is the one you are splitting (current).
+  </li>
+  <li>
+    The part on the right is the new part.
+  </li>
+  <li>
+    If current part is not last, the new part will be inserted between the current part and the next part.
+  </li>
+  <li>
+    If the current part is last, the new part will become last.
+  </li>
+</ul>
 
 <div style="width: 90vw; margin-left: calc(480px - 45vw)">
   <%= form_with model: @split_lesson_part_form, url: cip_year_module_lesson_part_split_path, method: :post do |f| %>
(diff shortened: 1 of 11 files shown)


app/views/core_induction_programme/lesson_parts/show_split.html.erb
```diff
@@ -0,0 +1,57 @@
+<% content_for :outside_container do %>
+<div class="govuk-width-container govuk-main-wrapper govuk-body">
+  <%= govuk_breadcrumbs breadcrumbs: course_lesson_breadcrumbs(@current_user, @course_lesson_part.course_lesson) %>
+  <h1 class="govuk-heading-l">Split lesson part</h1>
+  <ul>
+    <li>
+      The part on the left hand side is the one you are splitting (current).
```
callumacrae, line 7: this is very very inaccessible

kicferk1: Eh, it's meant to be internal. 
You mean the list is inaccessible? Or left/right distinction?

callumacrae: left/right distinction! both the words and the actual layout

kicferk1: Well, truth be told I think there is going to be one unlucky user of that page, which is Mark A. 

I don't think there is a very accessible way of handling this problem to begin with. The problem being 'splitting very long markdown roughly in half, in a text area, and showing what the outcome will look like'.

The amount of content to show is massive, so much so that showing it in a column, one after another, looked really unintuitive. That's why I opted to use visuals to show them in parallel.

I can rephrase it to be "The first part is the one you are splitting, the second part is new" or something like that if you think it's much better.

callumacrae: if it's for internal usage only I'm not sure it matters that much :)

amazimbe (approved):

rupollbabu (approved):

app/views/core_induction_programme/lesson_parts/show_split.html.erb
kicferk1: Alright :)

callumacrae (approved):

PR #14, opened 2021-02-25
Lesson part navigation

kicferk1:
### Context
We can split lesson parts, and so we can have multiple parts per lesson.
We want to see more than one!

1. Add navigation list to the top of the lesson part page
2. Add 'previous' and 'next' buttons at the bottom of lesson part page
3. Only show 'complete' option on the final lesson part
4. Sneakily replace all `link_to`s with `govuk_link_to`s. 

### Guidance to review
This is a purely frontend PR.
I made navigation elements into components, cause they look ugly and cause I think we might reuse them.
I stole the navigation look from https://www.gov.uk/universal-credit/eligibility, which in turn stole their css from [govuk publishing components](https://github.com/alphagov/govuk_publishing_components/blob/5d4ed0788b5dd2a9865b45eb7d062eb7f59dd73a/app/assets/stylesheets/govuk_publishing_components/components/_previous-and-next-navigation.scss). 
I renamed `gem-c-pagination` to `step-by-step-navigation`, not because I'm ashamed I stole from publishing components (hehe, govspeak), but because I think that's what we will use those styles for, and readability in our repo trumps consistent class names on the internet in my view.

### Testing
I haven't added any tests. Maybe I should (cypress?). Thought?


app/components/navigation_next_component.html.erb
```diff
@@ -0,0 +1,14 @@
+<li class="step-by-step-navigation__item step-by-step-navigation__item--previous">
+  <%= govuk_link_to(@url, class: "step-by-step-navigation__link") do %>
+    <span class="step-by-step-navigation__link-title">
+      <svg class="step-by-step-navigation__link-icon" xmlns="http://www.w3.org/2000/svg" height="13" width="17" viewBox="0 0 17 13">
```
seb-siem-softwire, line 4: Is there a better way of handling swg files - for example, https://stackoverflow.com/questions/36986925/how-do-i-display-svg-image-in-rails/36992492 provides some alternatives - there is a high possibility that we will be using more swg images and there is an opportunity here to create a more elegant solution

app/components/navigation_previous_component.html.erb
```diff
@@ -0,0 +1,14 @@
+<li class="step-by-step-navigation__item step-by-step-navigation__item--previous">
+  <%= govuk_link_to(@url, class: "step-by-step-navigation__link") do %>
+    <span class="step-by-step-navigation__link-title">
+      <svg class="step-by-step-navigation__link-icon" xmlns="http://www.w3.org/2000/svg" height="13" width="17" viewBox="0 0 17 13">
+        <path d="m6.5938-0.0078125-6.7266 6.7266 6.7441 6.4062 1.377-1.449-4.1856-3.9768h12.896v-2h-12.984l4.2931-4.293-1.414-1.414z"></path>
```
seb-siem-softwire, line 5: Similar to the above - is there a better way of handling svg files

app/views/core_induction_programme/lesson_parts/show.html.erb
```diff
@@ -31,5 +46,21 @@
       <% end %>
     <% end %>
 
+    <nav class="step-by-step-navigation" role="navigation">
+      <ul class="step-by-step-navigation__list">
+        <% if @course_lesson_part.previous_lesson_part %>
+          <%= render NavigationPreviousComponent.new(
```
seb-siem-softwire, line 52: Sweet!

app/components/navigation_next_component.html.erb
kicferk1: I'll look into this :)

rupollbabu (approved):
I have tested functionality, working as per Acceptance criteria.

app/components/navigation_next_component.html.erb
callumacrae: Or you can use webpack to add it as a background image in an SCSS file

kicferk1: Opted for the webpack option, cause we already had two images we stole from govuk publishing components there :)

app/views/layouts/application.html.erb
```diff
@@ -28,7 +28,7 @@
     <header class="govuk-header" role="banner" data-module="govuk-header">
       <div class="govuk-header__container govuk-width-container">
         <div class="govuk-header__logo">
-          <%= link_to "/", class: "govuk-header__link govuk-header__link--homepage" do %>
+          <%= govuk_link_to "/", class: "govuk-header__link govuk-header__link--homepage" do %>
             <span class="govuk-header__logotype">
```
seb-siem-softwire, line 32: Could you update this svg one as well

kicferk1: I'd rather not - it looks complicated (fallback images, what is this?), and more importantly, css for the classes that live on `application` svgs aren't in our code - they are from `govuk-frontend`. So we would have to add our versions of them to our code, with small differences, and then worry about maintaining them.

Also, I am quite happy keeping `application` file complicated and unreadable 😂

seb-siem-softwire: 👍

callumacrae (approved):

PR #20, opened 2021-03-02
Add current page user is visiting to breadcrumbs

xswoodsx:
### Context
The existing breadcrumbs don't include the page that the user is currently on or allow the user to return to the show page when editing using the breadcrumbs.

- updating breadcrumb helper to add breadcrumbs for course_year
- amending path of course_module and course_lesson breadcrumb 
- adding a method for the end crumb that shows either a link or string depending whether the user is on show or edit. As below 

### show - no link included in the end crumb
Screenshot 2021-03-02 at 13 49 55

### edit
Screenshot 2021-03-02 at 13 50 16

### Guidance to review
Had a weird problem with 2 tests failing on course lesson parts. It's not been replicated in github or on Michals laptop. It sorted itself once the db was dropped and recreated. If this happens on others devices it could need further looking at. 

### Testing
Testing of breadcrumb helper, including testing for behaviour with show and edit actions.

### DIFF
--- a/spec/helpers/cip_breadcrumb_helper_spec.rb
+++ b/spec/helpers/cip_breadcrumb_helper_spec.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe CipBreadcrumbHelper do
+  describe "core-induction-programme" do
+    let(:user) { create(:user, :admin) }
+    let(:course_year) { create(:course_year) }
+    let(:course_module) { create(:course_module, course_year: course_year) }
+    let(:course_lesson) { create(:course_lesson, course_module: course_module) }
+
+    it "returns an array for the course year breadcrumb" do
+      course_year_breadcrumb = helper.course_year_breadcrumbs(user, course_year)
+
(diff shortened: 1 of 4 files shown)


kicferk1 (commented):
Looks good to me. If you want to add some tests, we could unit test the breadcrumb helper.

Like, test the crumbs for a dummy year, module, lesson. Maybe test the different behaviour for different actions.

spec/helpers/cip_breadcrumb_helper_spec.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe CipBreadcrumbHelper do
+  describe "core-induction-programme" do
```
kicferk1, line 6: I was thinking more like 
```
RSpec.describe CipBreadcrumbHelper, type: :helper do
  describe "#year_breadcrumb" do
    ...
  end

  describe "#module_breadcrumb" do
    ...
  end

  describe "#lesson_breadcrumb" do
    ...
  end
end
```
And each describe can have a test for 'edit' action and a test for 'show' action.

xswoodsx: No problem, will add the others :)

kicferk1 (approved):

PR #22, opened 2021-03-03
Add completion_time_in_minutes to our seed files

kicferk1:
### Context
We added a new field (`completion_time_in_minutes`) to the db. We forgot to add it to our seed files. Nothing major, but we need to do it. 

As per title.

### Guidance to review
It is adding a field and a bunch of nils to both files. Github should be able to show you the diff if you want to read through it.


amazimbe:
I can't actually view the differences as the files are just too large. All I'm able to view is the raw file.

kicferk1:
> I can't actually view the differences as the files are just too large. All I'm able to view is the raw file.

What the hell, github shows the diff to me when I click to open it. Arghhh.

I guess you can check out the branch and compare it using a git client locally?

xswoodsx (approved):
I could view the differences fine, LGTM :)

seb-siem-softwire (approved):
👍

PR #23, opened 2021-03-03
Add test for govspeak accessibility

callumacrae:
Govspeak PR fixing accessibility issues: https://github.com/DFE-Digital/ecf-govspeak/pull/10

This PR will fail until that is merged.


kicferk1 (approved):
Looks good. I think you will need to update gemfile version of govspeak, and bundle install before it passes :)

kicferk1 (approved):

PR #26, opened 2021-03-05
Lesson part deletion

xswoodsx:
### Context
Admin can split out a lesson in to parts. We want them to also be able to delete If they can add, we want them to be able to delete parts

- Updated routes 
- Functionality for admin to delete a course lesson part, providing additional step to confirm. 
- Update testing for requests to delete and view show_delete page 
- Updating CourseLessonPart model and tests
- amending it block statements for non-admin user requests, it didn't reflect what was being tested - probably my bad from a previous commit. 

### Guidance to review
I added dependant destroy to the model during testing as it was raising a violation of foreign key constraint on update or delete. Without this, the delete would still work in the app, but not in testing. Probably be helpful to discuss why this was happening. 

There's no check to see if there's only one lesson part and not then letting admin delete that lesson.  

### Testing
Updated request and model tests added.

### DIFF
--- a/app/models/course_lesson_part.rb
+++ b/app/models/course_lesson_part.rb
@@ -3,7 +3,7 @@
 class CourseLessonPart < ApplicationRecord
   belongs_to :course_lesson
 
-  has_one :next_lesson_part, class_name: "CourseLessonPart", foreign_key: :previous_lesson_part_id
+  has_one :next_lesson_part, class_name: "CourseLessonPart", foreign_key: :previous_lesson_part_id, dependent: :destroy
   belongs_to :previous_lesson_part, class_name: "CourseLessonPart", inverse_of: :next_lesson_part, optional: true
 
   validates :title, presence: { message: "Enter a title" }, length: { maximum: 255 }
--- a/app/models/course_lesson_part.rb
+++ b/app/models/course_lesson_part.rb
@@ -3,7 +3,7 @@
 class CourseLessonPart < ApplicationRecord
   belongs_to :course_lesson
 
-  has_one :next_lesson_part, class_name: "CourseLessonPart", foreign_key: :previous_lesson_part_id, dependent: :destroy
+  has_one :next_lesson_part, class_name: "CourseLessonPart", foreign_key: :previous_lesson_part_id, dependent: :nullify
   belongs_to :previous_lesson_part, class_name: "CourseLessonPart", inverse_of: :next_lesson_part, optional: true
 
   validates :title, presence: { message: "Enter a title" }, length: { maximum: 255 }
--- a/app/policies/course_lesson_part_policy.rb
+++ b/app/policies/course_lesson_part_policy.rb
@@ -25,4 +25,12 @@ class CourseLessonPartPolicy < ApplicationPolicy
   def split?
     update?
   end
+
+  def show_delete?
+    update?
+  end
+
+  def destroy?
+    update?
+  end
 end
--- a/app/policies/course_lesson_part_policy.rb
+++ b/app/policies/course_lesson_part_policy.rb
@@ -27,10 +27,10 @@ class CourseLessonPartPolicy < ApplicationPolicy
   end
 
   def show_delete?
-    update?
+    destroy?
   end
 
   def destroy?
-    update?
+    update? && @record.course_lesson.course_lesson_parts.length > 1
   end
 end
--- a/app/views/core_induction_programme/lesson_parts/show_delete.html.erb
+++ b/app/views/core_induction_programme/lesson_parts/show_delete.html.erb
@@ -1,0 +1,17 @@
+<%= govuk_breadcrumbs breadcrumbs: course_lesson_breadcrumbs(@current_user, @course_lesson_part.course_lesson) %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <h2>Confirm you would like to remove <i><%= @course_lesson_part.title %></i> from <%= @course_lesson_part.course_lesson.title %>.</h2>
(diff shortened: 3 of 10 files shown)


app/models/course_lesson_part.rb
```diff
@@ -3,7 +3,7 @@
 class CourseLessonPart < ApplicationRecord
   belongs_to :course_lesson
 
-  has_one :next_lesson_part, class_name: "CourseLessonPart", foreign_key: :previous_lesson_part_id
+  has_one :next_lesson_part, class_name: "CourseLessonPart", foreign_key: :previous_lesson_part_id, dependent: :destroy
```
kicferk1, line 6: I am certain we don't want that 😂 That would wipe all the previous lesson parts in sequence.

`dependent : :nullify` is more like it.

I think in the app we were deleting second lesson part, and in tests you were deleting first one. difference is, if you delete first one, the second lesson part still has the id pointing at now gone first part.

spec/requests/core_induction_programme/course_lesson_part_spec.rb
```diff
@@ -53,6 +53,26 @@
...
+            content: "Updated content",
+            new_title: "Title two",
+            new_content: "Content two",
+          },
+        }
+        delete course_lesson_part_url, params: { id: course_lesson.course_lesson_parts[1].id }
+
+        course_lesson_url = "/core-induction-programme/years/#{course_year.id}/modules/#{course_module.id}/lessons/#{course_lesson.id}"
+
+        expect(CourseLessonPart.count).to eq(0)
```
kicferk1, line 71: That's the fault of destroy - but we are actually expecting one to be left :)

Cause we have two parts from the setup, and are only deleting one.

app/models/course_lesson_part.rb
xswoodsx: ahhh ok, yeah i saw nullify as well but thought destroy was getting rid of any reference to the deleted part rather than all the lesson parts.

callumacrae (approved):

app/views/core_induction_programme/lesson_parts/show_delete.html.erb
```diff
@@ -0,0 +1,17 @@
+<%= govuk_breadcrumbs breadcrumbs: course_lesson_breadcrumbs(@current_user, @course_lesson_part.course_lesson) %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <h2>Confirm you would like to remove <i><%= @course_lesson_part.title %></i> from <%= @course_lesson_part.course_lesson.title %>.</h2>
+      <div class="govuk-warning-text">
+        <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+        <div class="govuk-warning-text__text">This lesson part can not be retrieved once it has been deleted.</div>
+      </div>
+    <p>You can return to <b><%= @course_lesson_part.course_lesson.title %></b> without making any changes by clicking cancel</p>
```
kicferk1, line 9: Something something `div` has wrong indentation :)

app/policies/course_lesson_part_policy.rb
```diff
@@ -25,4 +25,12 @@ def show_split?
...
     update?
   end
+
+  def show_delete?
+    update?
+  end
+
+  def destroy?
+    update?
+  end
```
kicferk1, line 35: `show_delete` should probably be the same as `destroy`, rather than `update`. 

And, more importantly, we probably don't want to let the user delete the last lesson part - that would lead to some bad times. In here we should check that the `@record.course_lesson.course_lesson_parts.length > 1` or something like that, and if not, then return false.

app/views/core_induction_programme/lesson_parts/show_delete.html.erb
callumacrae: Any particular reason you're using `<b>` over `<strong>`?

xswoodsx: I forgot about strong to be honest but reading about differences <b> is used for drawing attention and with <strong> its to indicate content is of greater importance. I think <b> would be more  as it's just trying to clear what lesson it is the user is working on. 

Happy to go with either though, i didn't consider between the two initially.

spec/policies/course_lesson_part_policy_spec.rb
```diff
@@ -12,8 +12,18 @@
...
+
+    it "permits the show_delete action for admin when there is > 1 course lesson part" do
+      FactoryBot.create(:course_lesson_part, course_lesson: course_lesson_part.course_lesson)
+      is_expected.to permit_action(:show_delete)
+    end
+
+    it "permits the destroy action for admin when there is > 1 course lesson part" do
+      FactoryBot.create(:course_lesson_part, course_lesson: course_lesson_part.course_lesson)
+      is_expected.to permit_action(:destroy)
+    end
```
kicferk1, line 26: what about when there isn't more than 1 lesson part? :)

kicferk1 (approved):
Policy should probably have tests showing what happens when there is exactly 1 lesson part on a lesson, other than that looks good to me.

spec/policies/course_lesson_part_policy_spec.rb
xswoodsx: Would the tests on lines 15 & 16? cover that. I've not explicit said in the it statement, it's just that it forbids it. Happy to put it in an it block with a description if that'd make it easier to read.

kicferk1: Ah, you are right, they technically cover it.

Yes, I think a description would be very useful.

kicferk1 (approved):

PR #30, opened 2021-03-08
Update seed content from wordpress

kicferk1:
### Context
A lovely content person has been working on CIP content on wordpress for last two-three months. Before we do any massive content edits, we should get those in, because those will likely be much bigger, and because it's a pain in the ass to re-import them (as I found out this morning).

1. Replace \r\n in modules with just \n, they were introduced by accident
2. Update lesson parts with a ton of fixes - typos, off by 1 tables, removed video pending prefaces, possibly some more things. Around 3 months of work by a poor content person.
3. Update all remaining lesson times from Mark A's work

### Guidance to review
This is actually such a large change that github won't let you see it. You need to check out this branch and look through a commit manually - not that I think you would want to.

The steps I did were: 
1. Go back in time a month (code time travel, not real one sadly), when our data model was the same. 
2. Re-run wordpress script producing markdown that we used last time, on new export files
3. Copy-paste output of that into our seed file, run seeding fixing small definition ordering issues until it worked (just like last time)
4. Check out new code version, migrate the db with new data into current schema
5. Seed dump data into new schema
6. Take data in arrays from both old and new seed dumps. Iterate with indexes, and replace content of old lesson parts with content of new lesson parts. Load that into db.
7. Once it's all loaded, use regex replacing in RubyMine to fix youtube embeddings - half of them seems not to have been imported right, and I was too far in when I noticed. 

I _think_ these steps ought to work. I guess test of the pudding is in the working PR app.


app/models/core_induction_programme_exporter.rb
```diff
@@ -9,7 +9,7 @@ def run
       import: true,
     )
 
-    years = CourseYear.order(:created_at)
+    years = CourseYear.order(:title)
```
amazimbe, line 12: Just wondering how this orders when titles can start with both lowercase and uppercase letters. Do we need to restrict case?

kicferk1: I think titles of course years won't change any time soon.

The purpose of ordering in this file is to have a set order in migrations file. I thought `created_at` would more or less do that, but it was a little brittle, so I switched to something more stable and predictable.

As for 'do we need to worry about lower/upper case?' - very unlikely.

seb-siem-softwire (approved):

PR #32, opened 2021-03-08
Minor course lesson fix

xswoodsx:
### Context
Mark A spotted some minor fixes on the app. These changes are to resolve those finds.
[lesson duration trello card](https://trello.com/c/DabC5mU9/192-do-not-show-0-minutes-if-the-time-of-lesson-completion-is-exactly-1-hour-2-hours
)
[styling task list item trello card](https://trello.com/c/iGAJaQe3/191-remove-indentation-on-the-left-hand-side-of-lesson-lists)
- Refactor of duration_in_minutes_in_words method so that it doesn't show 0 minutes
- Removing the padding from the list of topic materials 

Screenshot 2021-03-08 at 16 22 05
Screenshot 2021-03-08 at 16 20 48

### Testing
Additional tests added to check the human readable time is not showing 0 minutes


app/models/course_lesson.rb
```diff
@@ -24,16 +24,12 @@ def course_lesson_parts_in_order
...
-
-    if number_of_hours.positive?
-      "#{number_of_hours} #{hour_string} #{number_of_minutes} #{minute_string}"
-    else
-      "#{number_of_minutes} #{minute_string}"
-    end
+    [].tap { |lesson_duration_string|
+      lesson_duration_string << "#{hours} hour".pluralize(hours)       unless hours.zero?
+      lesson_duration_string << "#{minutes} minute".pluralize(minutes) unless minutes.zero?
+    }.join(" ")
```
kicferk1, line 33: `lesson_duration_string` is a lying variable name - really, it's `lesson_duration_parts`, it's an array, and you add one or two elements to it. Then you join them with " " as the join element.

I would rename it, and maybe rewriting it so it doesn't use the `tap` - I am not convinced it makes things any more readable.

kicferk1 (approved):

PR #33, opened 2021-03-09
Fix staging environment

kicferk1:
### Context
We want to log in as dummy users on staging - we use it for content editing


xswoodsx (approved):
LGTM :)

PR #40, opened 2021-03-16
Additional cypress tests

xswoodsx:
### Context
There needs to be a check on all course lessons to make sure they're accessible, following, adding youtube titles. The cypress test added does a check on every CL.

- Cypress test to check accessibility on every course lesson
- Updated seed and cip_seed_dump files - as of 16/03/2021

### Guidance to review
The test should only run locally when you add --env tags=checkCourseLessonsAccessibility. This is due to the amount of time taken to run the test (upwards of 10 minutes). I looked at various ways to run it conditionally and most involved an additional library -[cypress select]( https://www.npmjs.com/package/cypress-select-tests ) and [cypress skip on](https://github.com/cypress-io/cypress-skip-test). For the sake of one test, it seemed too much so have gone with checking for a tag. 

The cip seed is as of 16/03, it includes changes that Mark A has been making, youtube titles and the changes to the html that was causing some accessibility tests to fail. 

To catch the accessibility fails, it was just a case of going through to find what was causing the failures and making the changes on the staging env. 

### Testing

As above. The accessibility test has been ran against the cip_seed in this PR.


callumacrae (approved):

spec/cypress/integration/accessibility.js
```diff
@@ -76,4 +76,20 @@ describe("Accessibility", () => {
 
     cy.checkA11y();
   });
-});
+  // this test should only be ran locally due to the length of time taken to complete.
+  // to include it add '--env tags=checkCourseLessonsAccessibility'
+  // to the yarn cypress:open cmd
```
kicferk1, line 81: I recommend capitalising comments (to make the reader feel like you've put enough work into them that they are worth reading, human psychology is a truly horrible thing). 

Also, add a space before comment, so it's obvious it refers to the test below.

kicferk1 (commented):
Just to make very sure, we don't need to apply any of these changes on staging, right? They are here so that if some poor soul decides to run the test they don't have loads of problems, but this is a snapshot of what the staging environment was like at some point in time? 

If so, umm, you will probably be better off re-doing that to get rid of conflicts, rather than resolving them yourself.

xswoodsx:
> Just to make very sure, we don't need to apply any of these changes on staging, right? They are here so that if some poor soul decides to run the test they don't have loads of problems, but this is a snapshot of what the staging environment was like at some point in time?
> 
> If so, umm, you will probably be better off re-doing that to get rid of conflicts, rather than resolving them yourself.

That's right, we don't need to apply any of these changes on staging, the seed dump from there has been used.

kicferk1 (approved):

PR #41, opened 2021-03-19
Create additional years for a cip

xswoodsx:
### Context
We want admin to be able to create additional years, Year 1, Year 2 etc.

- create a migration that sets the core_induction_programme_id to null: false(We don't want to have course_years that aren't assigned to a cip). 
- updated routes and policies to allow admin to create a new course year, whilst restricting non-admin users. 
- updated views, including adding new.html.erb for creating a course year. 
- updated specs for the above

### Guidance to review
I had a look at various ways of doing the migration to add null: false. 2-3 different ways worked, change_column_null, add_column/remove_column and the up/down set in the migration. I went with this as my first attempt with change_change column didn't allow a rollback. Also adding then removing the same column just with null: false seemed unnecessary. The problem i've come across is that when you try to create a new cip year is that i can't get rid of the "must exist" warning. I've tried various ways to just have the select provider but can't quite get it. 

Note - This was fixed by simply moving the validates... presence { message: "Select a provider"} above the belongs_to. Otherwise the first error message on the belongs_to takes precedent over the validation message. This message is just Rails' default 'must enter'

### Testing
Various tests have been added but there could be some additional ones required.

### DIFF
--- a/app/controllers/core_induction_programmes/years_controller.rb
+++ b/app/controllers/core_induction_programmes/years_controller.rb
@@ -10,16 +10,18 @@ class CoreInductionProgrammes::YearsController < ApplicationController
   before_action :load_course_year, except: %i[new create]
 
   def new
-    authorize User
+    authorize CourseYear
     @core_induction_programmes = CoreInductionProgramme.all
     @course_year = CourseYear.new
   end
 
   def create
-    @course_year = CourseYear.new(params.require(:course_year).permit(
-      :title, :content
-    ).merge(is_year_one: false, core_induction_programme: find_core_induction_programme))
-    authorize @course_year
+    authorize CourseYear
+    @course_year = CourseYear.new(
+      course_year_params.merge(
+        is_year_one: false, core_induction_programme: find_core_induction_programme,
+      ),
+    )
 
     if @course_year.valid?
       @course_year.save!
@@ -52,10 +54,12 @@ private
   end
 
   def course_year_params
-    params.permit(:title, :content)
+    return {} unless params.key? :course_year
+
+    params.require(:course_year).permit(:title, :content)
   end
 
   def find_core_induction_programme
-    CoreInductionProgramme.find_by(id: params[:course_year][:core_induction_programme_id])
+    CoreInductionProgramme.find_by(id: params[:course_year][:core_induction_programme])
   end
 end
--- a/app/controllers/core_induction_programmes/years_controller.rb
+++ b/app/controllers/core_induction_programmes/years_controller.rb
@@ -7,7 +7,28 @@ class CoreInductionProgrammes::YearsController < ApplicationController
 
   after_action :verify_authorized
   before_action :authenticate_user!
-  before_action :load_course_year
+  before_action :load_course_year, except: %i[new create]
+
+  def new
+    authorize User
+    @core_induction_programmes = CoreInductionProgramme.all
+    @course_year = CourseYear.new
+  end
+
+  def create
+    @course_year = CourseYear.new(params.require(:course_year).permit(
+      :title, :content
+    ).merge(is_year_one: false, core_induction_programme: find_core_induction_programme))
+    authorize @course_year
+
+    if @course_year.valid?
+      @course_year.save!
+      redirect_to cip_index_path
+    else
+      @core_induction_programmes = CoreInductionProgramme.all
+      render action: "new"
+    end
+  end
 
   def edit; end
 
@@ -31,6 +52,10 @@ private
   end
 
   def course_year_params
-    params.permit(:content, :title)
+    params.permit(:title, :content)
+  end
+
+  def find_core_induction_programme
+    CoreInductionProgramme.find_by(id: params[:course_year][:core_induction_programme_id])
   end
 end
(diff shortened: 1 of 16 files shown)


app/models/course_year.rb
```diff
@@ -4,11 +4,12 @@ class CourseYear < ApplicationRecord
   include CourseLessonProgressHelper
   include OrderHelper
 
-  belongs_to :core_induction_programme, optional: true
+  belongs_to :core_induction_programme
   has_many :course_modules, dependent: :delete_all
 
   validates :title, presence: { message: "Enter a title" }, length: { maximum: 255 }
   validates :content, presence: { message: "Enter content" }, length: { maximum: 100_000 }
+  validates :core_induction_programme_id, presence: { message: "Select a provider" }
```
kicferk1, line 12: I don't think you need to validate the id, I think `validate : core_induction_programme, ...` will work

app/controllers/core_induction_programmes/years_controller.rb
```diff
@@ -7,7 +7,28 @@ class CoreInductionProgrammes::YearsController < ApplicationController
...
+
+  def new
+    authorize User
+    @core_induction_programmes = CoreInductionProgramme.all
+    @course_year = CourseYear.new
+  end
+
+  def create
+    @course_year = CourseYear.new(params.require(:course_year).permit(
+      :title, :content
```
kicferk1, line 20: It would be convenient to handle title and content params for a year the same for creating and updating them - you might need to change the form in the edit view to have `model` attribute for it to work.

kicferk1 (commented):
I think if you stop validating the presence of id, and start validating the presence of the core induction programme, your duplicate default message will disappear :)

app/policies/user_policy.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class UserPolicy < ApplicationPolicy
+  def new?
+    admin_only
+  end
+
```
kicferk1, line 7: I don't think we want to touch users in this PR :)

app/views/core_induction_programmes/years/new.html.erb
```diff
@@ -0,0 +1,18 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">Create an additional course year</h1>
+    <%= form_with model: @course_year, url: years_path, method: :post do |f| %>
+      <%= f.govuk_error_summary %>
+      <%= f.govuk_collection_radio_buttons(
+            :core_induction_programme, @core_induction_programmes, :id, :name,
+          legend: { text: "Which provider would you like to create an additional year for?" , class: "govuk-heading-m" }
+          ) %>
```
kicferk1, line 9: Let's have one parameter per line if we do multiline components

app/views/core_induction_programmes/years/new.html.erb
```diff
@@ -0,0 +1,18 @@
...
+      <%= f.govuk_text_field :title, label: { text: 'Course year title' } %>
+      <%= f.govuk_text_area :content,
+                            label: { text: 'Course year content' },
+                            hint: { text: 'This content will be included at the top of the course year' },
+                            rows: 9 %>
```
kicferk1, line 14: Let's use 
```
<%= f.govuk_text_area(
  :content,
  ...
  rows: 10)%>
```
syntax - so there is less whitespace.

spec/models/core_induction_programme.rb
```diff
@@ -8,4 +8,15 @@
...
     it { is_expected.to have_many(:early_career_teachers).through(:early_career_teacher_profiles) }
   end
+
+  describe "course_years" do
+    it "returns multiple course years for a single core_induction_programme" do
+      core_induction_programme = FactoryBot.create(:core_induction_programme)
+      course_year_one = FactoryBot.create(:course_year, core_induction_programme: core_induction_programme)
+      course_year_two = FactoryBot.create(:course_year, core_induction_programme: core_induction_programme)
+
+      expect(course_year_one[:core_induction_programme]).to eql(course_year_two[:core_induction_programme])
```
kicferk1, line 18: `course_year_one.core_induction_programme`? I don't see why we would access it like a hash.

spec/requests/core_induction_programme/course_year_spec.rb
```diff
@@ -73,3 +118,11 @@
     end
   end
 end
+
+private
+
+def create_course_year
+  post "/years", params: { course_year: { title: "Additional year title",
+                                          content: "Additional year content",
+                                          core_induction_programme: course_year.core_induction_programme[:id] } }
```
kicferk1, line 127: I wonder if ending a line before title produces something with less whitespace.

app/policies/user_policy.rb
xswoodsx: Yeah, think this was my misunderstanding with the authorise User in the new action

spec/models/core_induction_programme.rb
xswoodsx: My bad, that's me forgetting we can access with that because of associations

kicferk1 (approved):

PR #43, opened 2021-03-25
Updating and creating modules

xswoodsx:
### Context

We want DfE admin to be able to create a new module, giving them the option to select where this appears in the order of modules and in which term. DfE should also be able to add titles and content when creating one. 

- Updating course_module controller with new and create actions. 
- Updated policies to allow new action
- Add term_and_title method to course module model. I've added this in as there's no check to see if selecting the module you want to follow the newly created module would be logical i.e. You could still create a module with a term of Autumn but select it's next module to be a summer term module. Adding this gives the DfE admin a view of what term the current modules are in.
- Course module New view - image below
- Previous and Next navigation components on Course Module show
Screenshot 2021-03-25 at 15 55 19 (2)
 
### Guidance to review
When using the version locally i created a module and then went to edit it (it was also happening for existing modules). There was an error being raised that wasn't being picked up in the tests caused by the assigning attributes in the controller. I changed the course_module_params method to permit :course_year_id instead of just course_year and this seems to have fixed it.

Some of the ways i've got this working feel a bit messy, notably... creating @course_modules and course_years in new and create. Also in the edit view where course_years are found with shown by doing...        
@course_module.course_year.core_induction_programme.course_years,

### Testing

Additional tests have been added for the model and updated course module policy tests. 
There's also additional request tests.


app/controllers/core_induction_programmes/modules_controller.rb
```diff
@@ -7,10 +7,37 @@ class CoreInductionProgrammes::ModulesController < ApplicationController
...
-  before_action :load_course_module
+  before_action :load_course_module, except: %i[new create]
 
   def show; end
 
+  def new
+    authorize CourseModule
+    @course_module = CourseModule.new
+    @course_years = CourseYear.where(core_induction_programme_id: params[:cip_id])
+    @course_modules = CourseModule.where(course_year_id: @course_years.first[:id])
```
kicferk1, line 18: `CourseModule.where(course_year_id: @course_years.map(&:id))` will give you more than one year :)

app/controllers/core_induction_programmes/modules_controller.rb
```diff
@@ -33,6 +60,18 @@ def load_course_module
...
+    params.require(:course_module).permit(:content, :title, :term, :course_year_id)
+  end
+
+  def find_course_year
+    CourseYear.find_by(id: params[:course_module][:course_year])
+  end
+
+  def set_previous_module(previous_module)
+    CourseModule.where(id: previous_module).pick(:previous_module_id)
+  end
```
kicferk1, line 72: So if we have module `A`, and another called `C`, and we insert `B` into the mix, what do we do?

We select `A` as previous module. But in here, we select the module _before_ A, which doesn't exist. So do we end up with B-A-C order?

app/views/core_induction_programmes/modules/edit.html.erb
```diff
@@ -4,6 +4,14 @@
...
     <%= form_with model:@course_module, url: year_module_path, method: :put do |f| %>
       <%= f.govuk_text_field :title, label: { text: "Module title" } %>
+      <%= f.govuk_collection_select(
+            :course_year,
+            @course_module.course_year.core_induction_programme.course_years,
+            :id,
+            :title,
+            label: { text: "Which year would you like to add a module to for?", class: "govuk-heading-m" },
+            hint: { text: "Course Years are listed in order" }
+          ) %>
```
kicferk1, line 14: Don't think we are adding any modules in edit.

Also, would be nice to have an option to set previous module here too.

app/views/core_induction_programmes/modules/new.html.erb
```diff
@@ -0,0 +1,37 @@
...
+            label: { text: "Which year would you like to add a module to for?" , class: "govuk-heading-m" },
+            hint: { text: "Course Years are listed in order" }
+          ) %>
+      <%= f.govuk_collection_select(
+            :previous_module_id,
+            @course_modules,
+            :id,
+            :term_and_title,
+            label: { text: "Where would you like the module to be placed?" , class: "govuk-heading-m" },
+            hint: { text: "The module you are creating will precede the module you are selecting" }
```
kicferk1, line 20: I think this line makes things harder for you.

`The module you are creating will follow the module you are selecting` makes for a much easier logic.

app/views/core_induction_programmes/modules/show.html.erb
```diff
@@ -28,5 +28,20 @@
         </ul>
       </li>
     </ol>
+    <nav class="step-by-step-navigation" role="navigation" aria-label="Lesson Part Navigation">
+      <ul class="step-by-step-navigation__list">
+        <% if @course_module.previous_module %>
+          <%= render NavigationPreviousComponent.new(
+            url: year_module_path(id: @course_module.previous_module.id),
```
kicferk1, line 35: Does `year_module_path(course_module.previous_module)` work?

app/views/core_induction_programmes/show.html.erb
```diff
@@ -1,6 +1,11 @@
 <%= govuk_breadcrumbs breadcrumbs: programme_breadcrumbs(@current_user, @core_induction_programme) %>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
+
+    <% if @current_user&.admin? %>
+      <%= govuk_link_to "Create CIP Module", cip_create_module_path(@core_induction_programme.id), button: true %>
```
kicferk1, line 6: does `cip_create_module_path` without any arguments work?

app/views/core_induction_programmes/modules/edit.html.erb
xswoodsx: Yeah, i thought about option for selecting previous module after sorting but definitely can add that in. 

My bad, the label text needs amending on this. It shouldn't say add, select the year you want to assign the module to or along those lines.

app/controllers/core_induction_programmes/modules_controller.rb
xswoodsx: So the user selects the module they want to follow the one they're creating. We're not selecting A as the previous for example, it's that A will follow it. I started off by selecting the previous module but changed because the one you want to follow has the existing previous module ID and will be the one you want to assign your new module to. Also it means in the select we don't have to include an additional option of say "Put this number one". It can be number one by selecting the first module. On the collection_select i've got a hint that says `"hint: { text: "The module you are creating will precede the module you are selecting" }`

app/views/core_induction_programmes/modules/new.html.erb
xswoodsx: Same as response above in modules controller. I did start doing it this way but switched.

app/views/core_induction_programmes/show.html.erb
xswoodsx: no, i get `missing required keys: [:cip_id]` error.

app/views/core_induction_programmes/modules/show.html.erb
xswoodsx: No, it just reloads the same page

app/views/core_induction_programmes/modules/new.html.erb
kicferk1: Soo, I think you are going to have a much easier time with `edit` if you let the user select the module preceding the one you are creating. 

As a general principle, because it reflects the db structure better.

As a concrete example, because when you try adding it to `edit`, it will be really hard to show a correct pre-selected value. Also, if you select previous module, it's a lot easier to find the things you need to edit - after all, you have the id of the `previous_module`, and the only other place to change is `pevious_module.next_module`.

So you could have
```
new_module = ...
previous_module = CourseModule.find(params[:previous_module_id])
next_module = previous_module.next_module

next_module.previous_module = new_module
previous_module.next_module = new_module
```
Which seems a little simpler than going to 'previous previous module'. 

If our database stored `next_module_id` on a module, then your solution would be neater, but I arbitrarily chose to store `previous_module_id` :P

kicferk1 (approved):
It looks good for now, I think there are a few more improvements to do on it (like making sure all edited modules are visible, maybe some cleanup here and there), but they can be done in their own PRs.

PR #44, opened 2021-03-30
General cleanup

kicferk1:
### Context
Some things improved on R&P that I want here too

1. Update Ruby and node versions
2. Remove references to `@current_user` - I don't think we need it.
3. Whitespace

### Guidance to review
Commit-by-commit

### Testing
If it builds and tests pass I think we are fine.


xswoodsx:
Have we been able to remove the set_current_user before_action and `@current_user` because we've set up the policies and have profile types?

xswoodsx (approved):
LGTM

kicferk1:
> Have we been able to remove the set_current_user before_action and `@current_user` because we've set up the policies and have profile types?

Nah, we were able to remove it because `current_user` helper works everywhere - I think early on I got tricked by RubyMine into thinking it wouldn't.

PR #47, opened 2021-03-31
Release to staging, number 5

kicferk1:
### Context
We want year, module and lessons editing on staging, so we can add more modules and years for our term structure :)


xswoodsx (approved):
All good! :)

PR #49, opened 2021-04-06
Migrations to set course years

xswoodsx:
### Context
We want to have course years set on the cip so that there is only 1 course year one and 1 course year two. 

adding migrations that 
- course year one and course year two references on a cip
- assign course year one and course year two using up and down methods
- remove is_year_one column from course years

updated spec tests

### Guidance to review
Should we update the views for assigning as year one or two in this PR or a follow up one?

### Testing

Updated spec/model and spec/factory for CY. Updated spec/model for core_induction_programme


db/migrate/20210401144737_set_first_and_second_course_years_for_core_induction_programme.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class SetFirstAndSecondCourseYearsForCoreInductionProgramme < ActiveRecord::Migration[6.1]
+  def up
+    CourseYear.all.each do |course_year|
+      cip = CoreInductionProgramme.find(course_year[:core_induction_programme_id])
+      if course_year.is_year_one?
+        cip.update!(course_year_one_id: course_year[:id])
+      else cip.update!(course_year_two_id: course_year[:id])
+      end
```
kicferk1, line 10: That's incomplete - it only ever updates one year. What if the programme you are migrating has two course years on it?

db/migrate/20210401144737_set_first_and_second_course_years_for_core_induction_programme.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def down
+    CoreInductionProgramme.all.each do |cip|
+      course_year = CourseYear.where(cip[:course_year_one_id]).or(CourseYear.where(cip[:course_year_two_id])).first
+      if cip[:course_year_one_id].nil?
+        course_year.update!(is_year_one: false)
+      else
+        course_year.update!(is_year_one: true)
+      end
+    end
```
kicferk1, line 22: This is incomplete too. You need to consider the case when both years are present :)

db/migrate/20210401144737_set_first_and_second_course_years_for_core_induction_programme.rb
xswoodsx: Line 5 had CourseYears.all.each but have switched this to go through each core induction programme, to follow the down method.

kicferk1 (approved):

PR #51, opened 2021-04-07
Cleanup routes

kicferk1:
### Context
Our CIP urls were pretty damn long, with a lot of unused ids in them. They were pretty unwieldy.

1. Split `resources` in routes up.
2. Update helper names, and arguments passed to them.
3. Update urls and names in tests.
4. Clean up all of `id: xxx.id` in path definitions.

### Guidance to review
I think all of our pages have some request tests on them, and that would have caught crashes around incorrect helpers.

We should probably make use of some `member`s in routes file, and enforce kebab-case, but that can probably come in a future PR.

### Testing
N/A


config/routes.rb
```diff
@@ -36,17 +36,16 @@
...
+  resources :years, controller: "core_induction_programmes/years", only: %i[new create edit update]
+  resources :modules, controller: "core_induction_programmes/modules", only: %i[show edit update]
+  resources :lessons, controller: "core_induction_programmes/lessons", only: %i[show edit update] do
+    resource :progress, controller: "core_induction_programmes/progress", only: %i[update]
+  end
+
+  resources :parts, controller: "core_induction_programmes/lesson_parts", only: %i[show edit update destroy] do
+    get "split", to: "core_induction_programmes/lesson_parts#show_split", as: "split"
+    post "split", to: "core_induction_programmes/lesson_parts#split"
+    get "show_delete", to: "core_induction_programmes/lesson_parts#show_delete"
```
ethanmills, line 48: Would it be worth adding a `scope path: "/", module: :core_induction_programmes` around all of this? Might not need to specify so many controllers

package.json
```diff
@@ -30,7 +30,7 @@
   "scripts": {
     "test": "jest --passWithNoTests",
     "cypress:open": "cypress open --project ./spec",
-    "cypress:run": "cypress run --project ./spec",
+    "cypress:run": "cypress run --project ./spec --config video=false,screenshotOnRunFailure=false",
```
ethanmills, line 33: Turns out this doesn't do anything, because the github action doesn't run this. I'll fix it soon, we need to change it in the config file

ethanmills (approved):
Looks much nicer 🙂 Just one note and a comment you're welcome to ignore

amazimbe (approved):

PR #52, opened 2021-04-08
Give users the ability to set a username

callumacrae:
### Context

https://dfedigital.atlassian.net/browse/CPDEL-212

Also pulls in some cucumber-related changes from the other repo

### Testing

yes

### DIFF
--- a/app/controllers/username_controller.rb
+++ b/app/controllers/username_controller.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class UsernameController < ApplicationController
+  before_action :authenticate_user!
+
+  def edit; end
+
+  def update
+    current_user.username = params[:user][:username]
+    current_user.save!
+
+    redirect_to dashboard_path
+  end
+end
--- a/app/controllers/username_controller.rb
+++ b/app/controllers/username_controller.rb
@@ -6,9 +6,13 @@ class UsernameController < ApplicationController
   def edit; end
 
   def update
-    current_user.username = params[:user][:username]
-    current_user.save!
-
+    current_user.update!(user_params)
     redirect_to dashboard_path
+  end
+
+  private
+
+  def user_params
+    params.require(:user).permit(:username)
   end
 end
--- a/app/controllers/username_controller.rb
+++ b/app/controllers/username_controller.rb
@@ -10,7 +10,7 @@ class UsernameController < ApplicationController
     redirect_to dashboard_path
   end
 
-  private
+private
 
   def user_params
     params.require(:user).permit(:username)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -37,6 +37,6 @@ class User < ApplicationRecord
   end
 
   def preferred_name
-    username && username.presence || full_name
+    username&.presence || full_name
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -35,4 +35,8 @@ class User < ApplicationRecord
   def course_years
     core_induction_programme&.course_years || []
   end
+
+  def preferred_name
+    username && username.presence || full_name
+  end
 end
--- a/app/views/username/edit.html.erb
+++ b/app/views/username/edit.html.erb
@@ -1,0 +1,15 @@
+<% content_for :before_content, govuk_back_link(
+  text: "Back",
+  href: dashboard_path)
+%>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">Edit your username</h1>
+
+    <%= form_with model: current_user, url: username_path do |f| %>
+      <%= f.govuk_text_field :username, label: { text: "Username" }, width: "two-thirds" %>
+      <%= f.govuk_submit "Save" %>
+    <% end %>
+  </div>
+</div>
--- a/db/migrate/20210407113051_add_username_to_users.rb
+++ b/db/migrate/20210407113051_add_username_to_users.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddUsernameToUsers < ActiveRecord::Migration[6.1]
+  def change
+    add_column :users, :username, :string
+  end
+end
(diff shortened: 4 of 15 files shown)


app/models/user.rb
```diff
@@ -32,4 +32,8 @@ def early_career_teacher?
   def mentor?
     mentor_profile.present?
   end
+
+  def preferred_name
+    username && username.presence || full_name
```
callumacrae, line 37: Is there a nicer way to do this in Ruby? In JavaScript I'd just write `username || full_name`, but an empty string isn't falsey in ruby

app/controllers/username_controller.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class UsernameController < ApplicationController
+  before_action :authenticate_user!
+
+  def edit; end
+
+  def update
+    current_user.username = params[:user][:username]
+    current_user.save!
```
kicferk1, line 10:
```
current_user.update!(user_params)
```

together with a private method
```
def user_params
  params.require(:user).permit(:username)
end
```
is how the pattern we tend to use.

app/models/user.rb
kicferk1: `username&.presence || full_name` would be shorter I guess.

app/views/username/edit.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  text: "Back",
+  href: dashboard_path)
+%>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">Edit your username</h1>
+
+    <%= form_with model: current_user, url: username_path do |f| %>
+      <%= f.govuk_text_field :username, label: { text: "Username" }, width: "two-thirds" %>
```
kicferk1, line 11: Do we need label text here? Seeing how it's similar to the name, maybe the gem does it for us.

kicferk1 (commented):
Looks good, tiny comments

app/controllers/username_controller.rb
kicferk1: Oh, also, do we want to validate it? In particular do we care about maximal length? Something to check with Isabelle. 

Then we would add validation to user model, check if the `update` without exclamation mark succeeds, and if not, we would render `edit`.

app/controllers/username_controller.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class UsernameController < ApplicationController
```
amazimbe, line 3: I don't have the full context but wouldn't it be better to have a user end point rather than a username endpoint. We can restrict the attributes that can be updated to just username if that's what is desired.

app/models/user.rb
amazimbe: The presence is not needed. username || full_name will work just fine

db/migrate/20210407113051_add_username_to_users.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddUsernameToUsers < ActiveRecord::Migration[6.1]
+  def change
+    add_column :users, :username, :string
```
amazimbe, line 5: You'll need to index this

kicferk1: Why would we need an index here? Do we expect to get users by their username?

amazimbe: Do we have a requirement regarding username uniqueness? It would be unusual if the usernames were not unique. But I take your point, we are not currently looking up the current user by username.

kicferk1: I think this is meant to be a 'preferred name', so I think we won't worry about uniqueness.

Many people would like to be referred to as 'Joe'.

Maybe we should call it `preferred_name` in the database, and `displayed_name` in the model method.

app/models/user.rb
callumacrae: `username || full_name` evaluated to an empty string when `username` is an empty string

i'll go with what @kicferk1 said, thanks!

app/views/username/edit.html.erb
callumacrae: We could make the title the label - I was just following the designs as I don't really have a strong opinion about this

kicferk1: I mean, does 
```
      <%= f.govuk_text_field :username, width: "two-thirds" %>
```
produce the same result as current code?

db/migrate/20210407113051_add_username_to_users.rb
callumacrae: I don't think it should be called username _anywhere_ (it's causing confusion already and we're the ones developing it), but I think if it's called username for the client it should be called username internally too.

If it's too confusing for us, we shouldn't be using that term for users either.

app/views/username/edit.html.erb
callumacrae: Ah sorry I misunderstood your original comment - I didn't realise the gem fills it in for us.

I think it's better to be verbose, especially if we end up changing what we call it internally (or externally) to preferred name or something like that

app/controllers/username_controller.rb
callumacrae: We don't really have the full context yet either - right now the username is the only thing that can be updated, but in the future if other stuff can be updated we don't know if it would happen on the same page or other pages.

I decided to go with a username controller instead of a more generic user controller as it's only one field that can be updated on this page, and it redirects to the dashboard instead of rendering the edit page again - I feel like with changing email or something like that you'd almost certainly want a different behaviour, and thus a different update method.

I don't think it'd be too tricky to change in the future, though - this isn't exactly a page that people will be bookmarking!

kicferk1 (approved):
Looks good to me, we can always change the `username` to something else later when our designers choose to label it differently.

app/controllers/username_controller.rb
callumacrae: Clarified with Isabelle, no validation for now.

PR #53, opened 2021-04-09
Restrict cip access for ect

xswoodsx:
### Context
We want to restrict access for ECTs so that they only have access to the cip they are enrolled on. 

- Implement CoreInductionProgrammePolicy, restricting index to admin and show to only the cip the ect is enrolled on
- Additional test user ECT profiles. Each CIP has an ECT profile assigned to it.
- Logout now redirects to sign_in page.

### Guidance to review
The logout redirects to sign_in because we've restricted the index to admin only. root_url is also the index page so we may need to think about what the landing page is. 

### Testing
Additional tests have been included for the CoreInductionProgrammePolicy and requests.

### DIFF
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -6,10 +6,10 @@ class Users::SessionsController < Devise::SessionsController
 
   TEST_USERS = %w[
     admin@example.com
-    early-career-teacher-1@example.com
-    early-career-teacher-2@example.com
-    early-career-teacher-3@example.com
-    early-career-teacher-4@example.com
+    ambition-institute-early-career-teacher@example.com
+    education-development-trust-early-career-teacher@example.com
+    teach-first-early-career-teacher@example.com
+    ucl-early-career-teacher@example.com
     mentor@example.com
   ].freeze
 
@@ -61,9 +61,5 @@ private
     user = User.find_by_email(email)
     sign_in(user, scope: :user)
     redirect_to dashboard_url
-  end
-
-  def after_sign_out_path_for(_resource_or_scope)
-    new_user_session_url
   end
 end
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -4,7 +4,14 @@ class Users::SessionsController < Devise::SessionsController
   class EmailNotFoundError < StandardError; end
   class LoginIncompleteError < StandardError; end
 
-  TEST_USERS = %w[admin@example.com early-career-teacher@example.com mentor@example.com].freeze
+  TEST_USERS = %w[
+    admin@example.com
+    early-career-teacher-1@example.com
+    early-career-teacher-2@example.com
+    early-career-teacher-3@example.com
+    early-career-teacher-4@example.com
+    mentor@example.com
+  ].freeze
 
   before_action :mock_login, only: :create, unless: -> { Rails.env.production? }
   before_action :redirect_to_dashboard, only: %i[sign_in_with_token redirect_from_magic_link]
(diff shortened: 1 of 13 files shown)


app/controllers/users/sessions_controller.rb
```diff
@@ -4,7 +4,14 @@ class Users::SessionsController < Devise::SessionsController
   class EmailNotFoundError < StandardError; end
   class LoginIncompleteError < StandardError; end
 
-  TEST_USERS = %w[admin@example.com early-career-teacher@example.com mentor@example.com].freeze
+  TEST_USERS = %w[
+    admin@example.com
+    early-career-teacher-1@example.com
```
kicferk1, line 9: Let's include the name of the course in the email please - remembering which course which number corresponds to is a pain.

app/controllers/users/sessions_controller.rb
```diff
@@ -55,4 +62,8 @@ def mock_login
     sign_in(user, scope: :user)
     redirect_to dashboard_url
   end
+
+  def after_sign_out_path_for(_resource_or_scope)
+    new_user_session_url
+  end
```
kicferk1, line 68: Let's add the start page the landing page and get rid of this :)

db/seeds/dummy_structures.rb
```diff
@@ -31,10 +31,12 @@
   end
 
   if EarlyCareerTeacherProfile.none?
-    user = User.find_or_create_by!(email: "early-career-teacher@example.com") do |u|
-      u.full_name = "ECT User"
-      u.confirmed_at = Time.zone.now.utc
+    CoreInductionProgramme.all.each_with_index do |cip, index|
+      user = User.find_or_create_by!(email: "early-career-teacher-#{index + 1}@example.com") do |u|
```
kicferk1, line 35: ah, that's what the indexes are for :/ 
yeah, I think we want course name in the dummy email, even if we repeat this code 4 times

kicferk1: Well, you could always add a helper method to commonise it.

spec/policies/core_induction_programme_policy_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  context "accessing cip views as admin" do
+    let(:user) { create(:user, :admin) }
+    it { is_expected.to permit_action(:show) }
+    it { is_expected.to permit_action(:index) }
+  end
+
+  context "accessing cip views as an early career teacher" do
+    let(:user) { create(:user, :early_career_teacher, { core_induction_programme: core_induction_programme }) }
+    it { is_expected.to permit_action(:show) }
+    it { is_expected.not_to permit_action(:index) }
```
kicferk1, line 18: I think you can write 
```
it { is_expected.not_to permit_action(:index) }
```
as
```
it { is_expected.to forbid_action(:index) }
```
I think it's slightly easier on the brain, but up to you if you want to write it this way :)

spec/requests/core_induction_programme/core_induction_programme_spec.rb
```diff
@@ -3,18 +3,56 @@
...
+        expect { get "/core-induction-programmes" }.to raise_error Pundit::NotAuthorizedError
+      end
+    end
+
+    describe "GET /core-induction-programmes/:id" do
+      it "renders the core_induction_programme show page" do
+        get "/core-induction-programmes/#{core_induction_programme.id}"
+        expect(response).to render_template(:show)
+      end
     end
```
kicferk1, line 46: Let's add a test for when there are two programmes, and teacher tries viewing the wrong one.

spec/requests/users/logout_spec.rb
```diff
@@ -17,7 +17,7 @@
 
       get "/users/sign_out"
 
-      expect(response).to redirect_to(root_path)
+      expect(response).to redirect_to(new_user_session_url)
```
kicferk1, line 20: yeah, let's keep root path here and add a start page as our root

kicferk1 (approved):

kicferk1:
At some point I will need to sit down and re-write our cypress tests in cucumber :P

PR #54, opened 2021-04-09
Reset db on re-deploy on dev

kicferk1:
### Context
So dev environment gets purged every now and then - to keep it from getting too messy, and to automatically bring staging changes into it. 

Add a rake task that only works in dev environments. The task deletes contents of our db and runs the seed files. 

Already done it on R&P.

### DIFF
--- a/lib/tasks/safe_reset.rake
+++ b/lib/tasks/safe_reset.rake
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+namespace :db do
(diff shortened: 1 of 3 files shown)


ethanmills (approved):
This looks very familiar

kicferk1:
> This looks very familiar

Yeah, but better - it won't crash the dev environment because I carefully specified which tables not to truncate.

PR #57, opened 2021-04-12
Add login process (sort of) for all users

callumacrae:
### Context

Moving to a login process that's like the old mock login process but for everyone

### Testing

i deleted them as we'd just be duplicating tests from inside Devise at this point. i figure if login breaks we'd notice pretty quickly during the review process anyway


lib/devise/strategies/yolo_authenticatable.rb
```diff
@@ -0,0 +1,24 @@
...
+
+require "devise/strategies/authenticatable"
+
+module Devise
+  module Strategies
+    class YoloAuthenticatable < Authenticatable
+      def authenticate!
+        if params[:user].present?
+          user = User.find_by(email: params[:user][:email])
+
```
kicferk1, line 11: Can you add a ticket to JIRA to sort our logging admins in a better way?

app/controllers/users/sessions_controller.rb
```diff
@@ -1,65 +1,3 @@
...
-  before_action :mock_login, only: :create, unless: -> { Rails.env.production? }
-  before_action :redirect_to_dashboard, only: %i[sign_in_with_token redirect_from_magic_link]
-  before_action :ensure_login_token_valid, only: %i[sign_in_with_token redirect_from_magic_link]
-
-  def create
-    super
-  rescue LoginIncompleteError
-    render :login_email_sent
-  rescue EmailNotFoundError
-    render :email_not_found
```
kicferk1, line 25: Those templates can go too. Also, I think cypress test is failing cause it tries to see text on one of them - time to update it :)

kicferk1 (approved):

lib/devise/strategies/yolo_authenticatable.rb
callumacrae: done https://dfedigital.atlassian.net/browse/CPDEL-258

PR #59, opened 2021-04-12
CPDEL-207: Confirm account

kicferk1:
### Context
We want to ask the users to create a username on their first login - or rather, on every login until they complete the flow. 
JIRA ticket: https://dfedigital.atlassian.net/browse/CPDEL-207

1. Add a flag to the user model checking whether they finished this lovely flow.
2. If yes, log them in as usual.
3. If not, log them in, and ask for username. 

### Guidance to review
I decided against using `confirmed` from devise - I think there is some magic surrounding it, and we will almost certainly end up discarding it.

Adding a flag to the db seemed like the easiest way of letting the users ignore username creation. If they want to they can press 'continue' and we will keep using their full name. And they will never see that flow again. 

### Testing
I added request specs.

Would be useful to check the behaviour matches expectations - we should have 4 fresh users to test with:
```
ambition-institute-early-career-teacher@example.com
education-development-trust-early-career-teacher@example.com
teach-first-early-career-teacher@example.com
ucl-early-career-teacher@example.com 
```

Cucumber tests will come in a later PR, when we cucumberise our setup. Soon, I promise.

### DIFF
--- a/app/views/username/new.html.erb
+++ b/app/views/username/new.html.erb
@@ -1,0 +1,15 @@
+<% content_for :before_content, govuk_back_link(
+  text: "Back",
+  href: dashboard_path)
+%>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">Create a username</h1>
+
+    <%= form_with model: current_user, method: :post, url: username_path do |f| %>
+      <%= f.govuk_text_field :username, label: { text: "Username" }, width: "two-thirds" %>
+      <%= f.govuk_submit "Save" %>
+    <% end %>
+  </div>
+</div>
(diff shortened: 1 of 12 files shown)


app/views/username/new.html.erb
```diff
@@ -0,0 +1,15 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">Create a username</h1>
+
+    <%= form_with model: current_user, method: :post, url: username_path do |f| %>
+      <%= f.govuk_text_field :username, label: { text: "Username" }, width: "two-thirds" %>
+      <%= f.govuk_submit "Continue" %>
+    <% end %>
+  </div>
+</div>
```
callumacrae, line 15: Is there any way to reuse the existing username template for this?

kicferk1: Ehh, I don't think it's worth the trouble. There are a few differences - text on submit button, header, the method that the form uses...

And the file is not very long either.

If we had more fields I would cut them out into their own partial, buuut for one field I am a bit too lazy.

callumacrae (approved):

PR #60, opened 2021-04-13
Update seed files

kicferk1:
### Context
So that dev has the latest version of staging

### Testing
All 4 cips should have their lessons split


xswoodsx (approved):
LGTM :)

amazimbe (approved):

PR #61, opened 2021-04-14
Upgrade release process

kicferk1:
### Context
I want to have the more mature deployment pipelines of R&P. 

Put all the relevant logic for each type of release into their own file - no reusing flows, because they are different.

Staging releases can (and should) be triggered via actions

Also added a way to fire off jobs in dev environment from github actions. 

### Testing
PR app working on this PR is test 1, test 2 is dev deployment working, test 3 is staging release working.


goodviber (approved):
LGTM

PR #62, opened 2021-04-15
Update login flow

kicferk1:
### Context
There are some improvements we need to make to our login flow - a few copy changes, and showing the error message on the field (rather than just on the summary).
Tickets:
1. https://dfedigital.atlassian.net/jira/software/projects/CPDEL/boards/89?selectedIssue=CPDEL-275
2. https://dfedigital.atlassian.net/jira/software/projects/CPDEL/boards/89?selectedIssue=CPDEL-277

1. Change copy on login flow
2. Sneakily add authenticating user before they try to see CIP page
3. Fix a bug with user never seeing the navigation if they somehow navigate out of creating username
4. Remove unused devise routes
5. Remove unused devise fields and concerns on user model
6. Override the default devise auth flow to make our form display errors correctly. 

### Guidance to review
I tried many ways of making warden happy, but no combination of `fail!`s seemed to work. So I always `success!` user, but leave some errors on them for our controller to pick up. 

### Testing
Try signing in with wrong email. You should get an error message in summary and on the field.
Sign in with the correct email. It should work, and take you to either the dashboard or creating a username flow.


app/controllers/users/sessions_controller.rb
```diff
@@ -1,3 +1,13 @@
 # frozen_string_literal: true
 
-class Users::SessionsController < Devise::SessionsController; end
+class Users::SessionsController < Devise::SessionsController
+  def create
+    self.resource = warden.authenticate!(auth_options)
```
amazimbe, line 5: Just curious why we are overriding devise here.

app/models/user.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class User < ApplicationRecord
-  devise :registerable, :trackable, :confirmable, :yolo_authenticatable
+  devise :trackable, :yolo_authenticatable
```
amazimbe, line 4: Never got to ask last time: how does the yolo strategy work?

amazimbe (approved):
LGTM! Just a couple of questions to aid my own understanding more than anything.

app/models/user.rb
callumacrae: you enter an email address and it just lets you in without a password or magic link

app/controllers/users/sessions_controller.rb
kicferk1: Because Devise believes in a different way of surfacing errors than we do.

Devise likes to have global `errors`, which it wants to surface as a notice. 
We like to have errors on models that we made form with - so we can automatically link them to the right fields.
And unfortunately, devise doesn't put those errors on the `resource` :/ 

And this is the simplest hack I found to get the behaviour we want - to add an error to the `email` fiend on the user.

callumacrae (approved):

PR #65, opened 2021-04-15
Silly change to test github actions setup

kicferk1:


kicferk1:
Well, it looks like 'Clever fix one' worked. Ehh, good lesson to learn I guess - never set up model based constants in your factories.

PR #66, opened 2021-04-16
CPDEL-286: Update service name in user-facing places

kicferk1:
### Context
https://dfedigital.atlassian.net/browse/CPDEL-286
We want service name updated

Update service name in page title, headers, and in cookies

### Guidance to review
N/A

### Testing
Check if it looks alright in the review app

### DIFF
--- a/app/views/cookies/show.html.erb
+++ b/app/views/cookies/show.html.erb
@@ -17,10 +17,10 @@
 
     <h1 class="govuk-heading-l">Cookies</h1>
     <p>Cookies are small files saved on your phone, tablet or computer when you visit a website.</p>
-    <p>We use cookies to make the Early Career Framework (ECF) work and collect information about how you use our service.</p>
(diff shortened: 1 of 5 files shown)


app/views/cookies/show.html.erb
```diff
@@ -17,10 +17,10 @@
 
     <h1 class="govuk-heading-l">Cookies</h1>
     <p>Cookies are small files saved on your phone, tablet or computer when you visit a website.</p>
-    <p>We use cookies to make the Early Career Framework (ECF) work and collect information about how you use our service.</p>
+    <p>We use cookies to make the Support for early career teachers work and collect information about how you use our service.</p>
```
callumacrae, line 20: in the other repo they've just written "the service" everywhere but the first time - might be easier to do the same here

callumacrae (approved):

PR #73, opened 2021-04-21
Percy

callumacrae:
### Context

Same as on the other repo!


kicferk1 (commented):
Is there a way to check for the randomness madness before we merge it in? Granted, we use a lot less randomness in our factories, but you know, for our sanity's sake :D

callumacrae:
> Is there a way to check for the randomness madness before we merge it in? Granted, we use a lot less randomness in our factories, but you know, for our sanity's sake :D

I manually check it, seems fine to me

kicferk1 (approved):

PR #76, opened 2021-04-22
CPDEL-304: Mentor materials

kicferk1:
### Context
We will want some mentor materials in our app. There will be a lot of them, with a lot of content, so what better way to show them than by govspeak? 
https://dfedigital.atlassian.net/browse/CPDEL-304

Add mentor materials model, and a very basic show/index pages for them. Add them to our seed dump.

### Guidance to review
I don't really know what the mentor materials will point to and how they will be used, so I made them point at a lot of things, all optional. I hope the analysis of the structure that we are aiming for is going to finish soon, but until then, there isn't much I can do to make it better :/

### Testing
Just unit tests for time being - I am certain cucumber tests will come what we understand interaction with module materials better.


xswoodsx (approved):
All LGTM :)

PR #77, opened 2021-04-22
Add placeholder CIP guidance pages

callumacrae:
### Context

https://dfedigital.atlassian.net/browse/CPDEL-312

### Testing

yes


app/controllers/training_and_support_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class TrainingAndSupportController < ApplicationController
+  before_action :authenticate_user!
+
+  def show
+    @provider = current_user.core_induction_programme
+
+    if @provider.blank?
+      raise Pundit::NotAuthorizedError, "This page is only available to Early Career Teachers"
```
callumacrae, line 10: Not really sure what to do here but the only way it can be accessed is through the URL directly so I'm not sure it's worthy of a dedicated page

kicferk1: Well, we certainly don't have designs for that page, so this is good enough for now I think.

app/views/training_and_support/show.html.erb
```diff
@@ -0,0 +1,9 @@
+<% if @provider.name == "Ambition Institute" %>
+  <h1 class="govuk-heading-l">Guidance for Ambition Institute</h1>
+<% elsif @provider.name == "Education Development Trust" %>
+  <h1 class="govuk-heading-l">Guidance for Education Development Trust</h1>
+<% elsif @provider.name == "Teach First" %>
+  <h1 class="govuk-heading-l">Guidance for Teach First</h1>
+<% elsif @provider.name == "UCL" %>
+  <h1 class="govuk-heading-l">Guidance for UCL</h1>
+<% end %>
```
kicferk1, line 9: I know we don't have much in them, buuuuut...

Could you make a partial for each of them? And in this template, just render the correct partial, no contents? I think that will have to happen as soon as we start adding content anyway.

spec/cypress/app_commands/scenarios/ect_ucl.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+cip = FactoryBot.create(:core_induction_programme, name: "UCL")
+# You have to create this user in your spec before running this scenario
+User.find("53960d7f-1308-4de1-a56d-de03ea8e1d9a").core_induction_programme = cip
```
kicferk1, line 5: I do wonder if that's the correct approach. 

I am quite tempted to load our seed file into our test database and use the users and programmes we have in there. It should make our tests reflect the dev apps more closely, and it ought to make the repeatable CIP setup less frustrating (and closer to reality). 

The cost obviously being the time it takes to load our seed file.

Do you want to see whether that idea makes sense in reality?

kicferk1 (commented):
I think this approach satisfies the conditions on the ticket, and perhaps it will be enough. I am moderately worried that we will want to make these pages public at some point, and then this solution obviously can't work, but perhaps that's a problem we should solve when we get to it. 

I would love to see if we can use our seed in cucumber setup though - perhaps as a step "given the app was seeded with dev environment data", to see how slow it would be. Kind of how we do for the automated accessibility test of all lesson parts.

spec/cypress/app_commands/scenarios/ect_ucl.rb
callumacrae: There's two problems here:

- The way this is done now is a bit weird because the user was created in Cypress and then the scenario was ran afterwards - if we had the ability to log in a user created in a scenario we could merge all four of these scenarios into one and it would be a lot cleaner
- Yea, the lack of real data isn't ideal. We could give it a go, but I suspect a hybrid approach would be ideal - could we have a scenario which just loads the seed data?

kicferk1: I'm happy for loading of seed data to be in a scenario if it makes our lives easier.

kicferk1 (approved):

kicferk1:
```
  1) TrainingAndSupports GET /show returns http success
     Failure/Error: expect(response).to have_http_status(:success)
       expected the response to have a success status code (2xx) but it was 302
     # ./spec/requests/training_and_support_request_spec.rb:9:in `block (3 levels) in <top (required)>'
```
Just get rid of the success check, checking that a template is rendered is already excluding any exceptions (I think).

ethanmills (commented):
Cypress steps look great!

PR #78, opened 2021-04-26
Remove in-memory dictionaries for loading seed files

kicferk1:
### Context
Hope for the best. Just for code readability.


ethanmills (approved):
More than happy for you to try it out here first 😛

kicferk1:
> More than happy for you to try it out here first 😛

That would be the opposite to what we usually do xD

PR #79, opened 2021-04-26
CPDEL-314: Add basic dummy API client

kicferk1:
### Context
We will want to make a lovely API client that will fetch user data from R&P. There is a lot of features that it will have, but so far I settled for something really basic that adds the structure in place - so we can enhance it in small steps. 

I added a basic API client (that I copied from Apply for Teacher Training).

### Guidance to review
It's not actually used anywhere - I got the data on what the requests and responses look like by editing it a little. I _think_ the way it will work eventually is "fire a sidekiq job every 5 minutes that syncs our users", and "add a 'updated_since' parameter to our request". Probably will need some auth too, better error handling, etc. etc. 

Once it gets a response, it will update our data - not too much of it. The idea is that once we have a user, it's more or less our responsibility to keep their info up-to-date. Except things that we can't - maybe like their CIP or their connections to other users.

### Testing
No tests yet, because it doesn't do much.


.env.template
```diff
@@ -4,4 +4,5 @@ GOVUK_NOTIFY_API_KEY=
 DOMAIN=http://localhost:3000
 GOVUK_APP_DOMAIN=http://localhost:3000
 GOVUK_WEBSITE_ROOT=http://localhost:3000
-GIAS_API_ROOT=https://ea-edubase-api-prod.azurewebsites.net
\ No newline at end of file
+GIAS_API_ROOT=https://ea-edubase-api-prod.azurewebsites.net
+REGISTER_AND_PARTNER_URL=http://localhost:3001
```
callumacrae, line 8: pls tell your editor to start ending files with new lines (or, ending every line with an End Of Line character) lol

If you install editorconfig it does it for you: https://github.com/DFE-Digital/ecf-engage-and-learn/blob/develop/.editorconfig

ethanmills (approved):

PR #80, opened 2021-04-27
Tests for mentor login flow

callumacrae:
### Context

https://dfedigital.atlassian.net/browse/CPDEL-322
https://dfedigital.atlassian.net/browse/CPDEL-323
https://dfedigital.atlassian.net/browse/CPDEL-356

This mostly already existed, so this was mostly just improving the tests.

### Testing

yes


app/helpers/application_helper.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module ApplicationHelper
+  def service_name_for(current_user = nil)
+    return "Mentoring for early career teachers" if current_user&.mentor?
```
kicferk1, line 5: Let's name the variable `user` or something like that - `current_user` is also a helper method from devise, available in a lot of places, and I don't want to confuse the two.

spec/helpers/application_helper_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    it "returns correct service name if current user is ECT" do
+      mentor_user = create(:user, :early_career_teacher)
+      expect(helper.service_name_for(mentor_user)).to eq("Support for early career teachers")
+    end
+
+    it "returns correct service name if current user is induction coordinator" do
+      mentor_user = create(:user, :induction_coordinator)
+      expect(helper.service_name_for(mentor_user)).to eq("Support for early career teachers")
+    end
```
kicferk1, line 30: Funnily enough, induction coordinators can be mentors too 😭

kicferk1 (commented):
Looks pretty good. Did you track down what was causing the issue with factory bot not creating a useful mentor user?

spec/helpers/application_helper_spec.rb
callumacrae: luckily, this one isn't!!

callumacrae:
> Did you track down what was causing the issue with factory bot not creating a useful mentor user?

no idea lol

kicferk1 (approved):

kicferk1:
> > Did you track down what was causing the issue with factory bot not creating a useful mentor user?
> 
> no idea lol

Fine, in that case I will choose the answer I like. I blame vim, somehow it must be its fault 😂

callumacrae:
> Fine, in that case I will choose the answer I like. I blame vim, somehow it must be its fault 😂

Seems likely 👀

spec/helpers/application_helper_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+# Specs in this file have access to a helper object that includes
+# the ApplicationHelper. For example:
+#
+# describe ApplicationHelper do
+#   describe "string concat" do
+#     it "concats two strings with spaces" do
+#       expect(helper.concat_strings("this","that")).to eq("this that")
+#     end
+#   end
+# end
```
kicferk1, line 14: I just noticed, I think this is just a comment for any `:helper` rspec type, right? Can we cut it?

kicferk1 (commented):
Sooo, I know you spent a while implementing it, but do we want to remove the dynamic service name now?

callumacrae:
:lolsob:

PR #81, opened 2021-04-28
Different summaries for ECTs and mentors

kicferk1:
### Context
We will need different summaries on modules and lessons for ECTs and mentors.

Rename `content` on modules to `ect_summary`. Add `mentor_summary` to modules.
Add `ect_summary` and `module_summary` to lessons.

Fix all tests and displays.

### Guidance to review
This should probably go in after the PR adding mentor materials.


db/schema.rb
```diff
@@ -77,11 +79,12 @@
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
     t.string "title", null: false
-    t.text "content", null: false
+    t.text "ect_summary", null: false
```
xswoodsx, line 82: How come ECT summary is the only one with null false in both modules and lessons?

xswoodsx (approved):
Other than the one comment, mainly out of curiosity, everything looks good :)

db/schema.rb
kicferk1: Ehh, it's the only one that is currently guaranteed to be there.

The others will _eventually_ be not-null, but not yet.

PR #82, opened 2021-04-28
Add mentor materials content

xswoodsx:
### Context

We want to add mentor material documents to the application. As a starter we've added Ambition materials, primarily because they have the most like for like structure to the existing course_modules and course_lessons. The method for initially importing has been kept in dummy structures to use for the other cips. 

- Updated seed and updated seed dump. This has been taken from staging, with mentor materials seeded and dumped, with the dumped file being the latest version in the PR. 
- Removing the two-thirds div because the material was looking really compact.
- Commented out method for initially importing each of the material files. 

### Guidance to review
A full list of the Ambition materials can be found in the PR app - https://ecf-engage-and-learn-review-pr-82.london.cloudapps.digital/

### Testing
None added


app/views/core_induction_programmes/mentor_materials/show.html.erb
```diff
@@ -1,9 +1,6 @@
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
```
kicferk1, line 2: If you don't want `govuk-grid-column-two-thirds`, use `govuk-grid-column-full` :)

kicferk1 (approved):

PR #83, opened 2021-04-29
Better errors when login fails

callumacrae:
### Context

https://dfedigital.atlassian.net/browse/CPDEL-340

### Testing

yes

### DIFF
--- a/lib/devise/strategies/yolo_authenticatable.rb
+++ b/lib/devise/strategies/yolo_authenticatable.rb
@@ -21,7 +21,7 @@ module Devise
             return
           end
 
-          unless email.match(/.@./)
+          unless email.match(Devise.email_regexp)
             user = User.new
             user.errors.add :email, "Enter an email address in the correct format, like name@school.org"
             success! user
--- a/lib/devise/strategies/yolo_authenticatable.rb
+++ b/lib/devise/strategies/yolo_authenticatable.rb
@@ -5,14 +5,36 @@ require "devise/strategies/authenticatable"
 module Devise
   module Strategies
     class YoloAuthenticatable < Authenticatable
+      # Validation is handled by authenticate! below so that errors look nicer
+      def valid?
+        true
+      end
+
       def authenticate!
         if params[:user].present?
-          user = User.find_by(email: params[:user][:email])
+          email = params[:user][:email]
+
+          if email == ""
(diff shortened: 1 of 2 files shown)


lib/devise/strategies/yolo_authenticatable.rb
```diff
@@ -5,14 +5,36 @@
...
+          email = params[:user][:email]
+
+          if email == ""
+            user = User.new
+            user.errors.add :email, "Enter an email address"
+            success! user
+            return
+          end
+
+          unless email.match(/.@./)
```
kicferk1, line 24: I guess we don't want devise regex here?

kicferk1 (approved):

lib/devise/strategies/yolo_authenticatable.rb
callumacrae: I couldn't find it in a variable I could access (in fact, I couldn't find any information about it at all except the actual value posted on Stack Overflow…) so I just went with the simplest regex I could think of. Devise doesn't seem to check against a regex anyway so even though the regex I found was slightly more strict, that wouldn't cause any issues.

callumacrae: (deleted my previous comment for being lazy and wrong)

I can change it if required? This matches what was said in the ticket.

`Devise::email_regexp` isn't actually being used by Devise, though—it was perfectly happy letting stuff like `rwghiuhefw` though before this change—so this slightly less strict expression shouldn't cause any problems.

kicferk1: Does `Rails.application.config.email_regexp` refer to it? We have it in our devise initialiser.

kicferk1: I don't want to have multiple email regexes around the codebase, and it would be nice to have that in our application config, just in case we want to use it in many places.

callumacrae:
```js
config.email_regexp = /\A[^@\s]+@[^@\s]+\z/
```

Is that setting `Rails.apploication.config.email_regexp` or something inside Devise? That's just the regex that Devise already used lol

kicferk1 (approved):

PR #84, opened 2021-04-29
Add sending invite emails to ECTs

kicferk1:
### Context
We want to be able to send invitation emails to ECTs

1. Switch to use our own notify instance.
2. Add new templates
3. Remove old templates
4. Add single table inheritance for emails to track what we sent
5. Add some tests
6. Add a job to send emails to users

### Guidance to review
If you want to be invited to new notify instance, let me know.

### Testing
I should test sending one email via task in review app. Using rails console.


kicferk1:
Alright, I checked and after making myself a user and scheduling a task via cloudfoundry, I got an email.

callumacrae (approved):

PR #85, opened 2021-04-30
Editing mentor materials

kicferk1:
### Context
We want to be able to edit mentor materials.

Add editing mentor materials for admin users. Add needed policies.

### Guidance to review
This is for internal use only - so we can fix any issues we run into. Closely follows editing all other content-related things.

### Testing
Unit testing. I think it makes sense to hold off cucumber tests until we have concrete user stories for them.

If you want to see it in action, login as admin, go to `/mentor-materials` in the url, click on a mentor material and you should see a green 'edit' button.

### DIFF
--- a/spec/cypress/integration/CipStoriesAdmin.feature
+++ b/spec/cypress/integration/CipStoriesAdmin.feature
@@ -84,3 +84,26 @@ Feature: Admin user interaction with Core Induction Programme
     Then "page body" should contain "Your changes have been saved"
     And "page body" should contain "Lesson part test title"
     And "page body" should contain "New content for lesson part"
+
+  Scenario: Can edit mentor materials
+    Given mentor_material was created with id "a4dc302c-ab71-4d7b-a10a-3116a778e8d5"
+    And I am on "core induction programme mentor material" page with id "a4dc302c-ab71-4d7b-a10a-3116a778e8d5"
+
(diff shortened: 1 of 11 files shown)


callumacrae:
> I think it makes sense to hold off cucumber tests until we have concrete user stories for them.

Cucumber specs are also super useful when reviewing! 🙏

kicferk1:
> Cucumber specs are also super useful when reviewing! 🙏

Fine, I will add some :)

spec/cypress/integration/CipStoriesAdmin.feature
```diff
@@ -84,3 +84,26 @@ Feature: Admin user interaction with Core Induction Programme
     Then "page body" should contain "Your changes have been saved"
     And "page body" should contain "Lesson part test title"
     And "page body" should contain "New content for lesson part"
+
+  Scenario: Can edit mentor materials
```
callumacrae, line 88: I'm not really a fan of this being in this file, it feels like a bunch of basically unrelated features are ending up in this file purely because of who they're available to - but not sure if splitting it up is a task for this PR or another

kicferk1: Ehh, I think I prefer that to be another PR - cause it's not just this file. It's also `CIPStoriesECT` :/

PR #87, opened 2021-05-04
Changes to preferred name behaviour

callumacrae:
- Rename username to preferred name https://dfedigital.atlassian.net/browse/CPDEL-212
- Remove initial "set username" screen https://dfedigital.atlassian.net/browse/CPDEL-364
- Some other copy changes https://dfedigital.atlassian.net/browse/CPDEL-376

### Guidance to review

Per one of the tickets, the `account_created` logic is sticking around and might be used in the future. It was already being set to `true` in the dashboard controller which is the new desired behaviour.

### Testing

yes


db/seeds/cip_seed_dump.rb
```diff
@@ -704,3 +704,6 @@
...
 ])
+CoreInductionProgramme.import([:id, :name, :course_year_one_id, :course_year_two_id], [
+  ["57482e25-2654-4c54-9219-8583d3336d1e", "Test Core induction programme", nil, nil]
+])
```
kicferk1, line 709: Let's not merge this, I need to track why it's happening and fix it.

kicferk1 (approved):
Looks good to me - so long as we remove the change to the seed files, I need to get that sorted.

kicferk1:
I think
```
  Scenario: Should be able to load seed data
    Given seed data should be loaded
    And I am logged in as existing user with email "ambition-institute-early-career-teacher@example.com"
    When I click the submit button
    Then "page body" should contain "Ambition Institute ECT User"
```
is failing because it expects the create account screen too. If you remove the `When I click submit button` line it ought to just work.

PR #88, opened 2021-05-04
Use ect name, not preferred name

callumacrae:
### Context

i'm an idiot. merged prs in wrong order. this is the fix 😅


kicferk1 (approved):

kicferk1:
Happens to us all sometimes :D

PR #90, opened 2021-05-05
CPDEL-339: Update progress tracking

kicferk1:
### Context
https://dfedigital.atlassian.net/browse/CPDEL-339
We have details of behaviour around progress setting and displaying, so we need to update the old mechanisms around it. 

1. Wipe any existing status from progresses - we don't have live data yet. This is to set a different default on the enum field.
2. Update tag colours for progress states - remove one unnecessary progress
3. Add validation to the lesson progress form. This was a lot harder than I expected. I needed to move progress setting to lesson part controller - to conform to pattern of showing errors on the same page. In order to display error summary on top of a page I also had to do some gymnastics - separating lesson content into a partial.
4. Plenty of changes in tests, but mostly shuffling things around. Just like the cucumber change suggests.

### Guidance to review
Maybe I should move authorisation check from lesson policy to progress policy, it might make more sense there.

### Testing
In order to test it, go to review app, log in as one of the users, say `ucl-early-career-teacher@example.com`, go to your materials, go to a lesson, move to last part, and play with setting progress.


spec/cypress/integration/CipStoriesECT.feature
```diff
@@ -21,24 +21,22 @@ Feature: ECT user interaction with Core Induction Programme
...
     And I click the submit button
-    Then "tag component" should contain "in progress"
+    Then "tag component" should contain "to do"
+    And percy should be sent snapshot
 
     When I click on "link" containing "Test Course lesson"
     And I click on "I feel confident that I’ve read and understood this session" label
     And I click the submit button
     Then "tag component" should contain "complete"
+    And percy should be sent snapshot
```
callumacrae, line 42: I don't think it's worth sending two snapshots of the same page - plus it will fail anyway as it is because they have the same name

app/controllers/core_induction_programmes/lesson_parts_controller.rb
```diff
@@ -57,12 +57,36 @@ def destroy
...
+      load_progress
+    end
+  end
+
+  def load_progress
+    @lesson_progress = CourseLessonProgress.find_or_create_by!(
+      early_career_teacher_profile: current_user.early_career_teacher_profile,
+      course_lesson: @course_lesson_part.course_lesson,
+    )
+    @lesson_progress.progress = nil
```
asmega, line 89: why do we set this to `nil`? don't we want to load what was persisted in the database?

kicferk1: Yes, as far as I can tell. We don't really expect the user to give us the same answer many times, or re-visit the page once it was completed.

I guess maybe if the progress is 'complete' we would want to keep it, but that's probably something to do user research on.

app/controllers/core_induction_programmes/lesson_parts_controller.rb
```diff
@@ -72,4 +96,10 @@ def course_lesson_part_params
   def lesson_split_params
     params.require(:split_lesson_part_form).permit(:title, :content, :new_title, :new_content)
   end
+
+  def lesson_progress_params
+    return {} unless params.key?(:course_lesson_progress)
+
+    params.require(:course_lesson_progress).permit(:progress)
```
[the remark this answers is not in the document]
kicferk1: Cool, I wish I knew this trick sooner!

app/models/course_year.rb
```diff
@@ -59,8 +59,8 @@ def compute_user_course_module_progress(lessons_with_progresses, modules_in_orde
   end
 
   def course_module_progress_status(lessons)
```
asmega, line 61: not for this PR but we should move this to a service class later. `CourseYear` does not need to know anything about the progress. Also this method despite being in this class never actually references the `CourseYear` instance

asmega (approved):

PR #91, opened 2021-05-06
CIP changes from wireframes (minus test updates)

callumacrae:
### Context

wip but had to commit because this is already more than a days work!


app/views/core_induction_programmes/years/_module_list.html.erb
```diff
@@ -0,0 +1,22 @@
...
+        </h3>
+
+        <div class="gem-c-govspeak govuk-govspeak">
+          <%= course_module.ect_summary_to_html.html_safe %>
+        </div>
+
+        <div class="app-task-list__section">
+          <span><%= course_module.course_lessons.count.humanize.capitalize %> self-study sessions</span>
+          <%= render ProgressLabelComponent.new progress: course_module.progress %>
+        </div>
```
kicferk1, line 16: I would ask if we want to pluralise `sessions` but I don't think there is any module with a single session, sooo... we can probably let it slide.

app/views/core_induction_programmes/years/show.html.erb
```diff
@@ -0,0 +1,44 @@
...
+        <%= component.nav_item(path: year_path(@cip.course_year_one_id)) do %>
+          Year 1
+        <% end %>
+        <%= component.nav_item(path: year_path(@cip.course_year_two_id)) do %>
+          Year 2
+        <% end %>
+      <% end %>
+    <% end %>
+
+    <% if current_user&.admin? || policy(@course_year).edit? %>
```
kicferk1, line 18: Why would we need both an admin check and policy check?

app/views/core_induction_programmes/years/show.html.erb
```diff
@@ -0,0 +1,44 @@
...
+      <div class="govuk-button-group">
+        <% if current_user&.admin? %>
+          <%= govuk_link_to "Create CIP Module", cip_create_module_path(@cip), button: true %>
+          <%= govuk_link_to "Create CIP Lesson", cip_create_lesson_path(@cip), button: true %>
+        <% end %>
+        <% if policy(@course_year).edit? %>
+          <%= govuk_link_to "Edit year content", edit_year_path(@course_year), button: true %>
+        <% end %>
```
kicferk1, line 26: I think for time being we can lump it all behind a check 'is user an admin?', they are the only ones who can edit content for now.

kicferk1 (commented):
Looks pretty good, tiny comments

app/views/core_induction_programmes/years/show.html.erb
callumacrae: This code was moved from c_i_p/show.html.erb - I didn't write it

callumacrae: oh wait i totally did write this

i was just combining the two existing if statements to see if there should be a button group - i didn't really think about it that hard

PR #92, opened 2021-05-07
CPDEL-392 Can add content-less lessons

asmega:


app/models/core_induction_programme.rb
```diff
@@ -1,15 +1,14 @@
...
 class CoreInductionProgramme < ApplicationRecord
-  has_many :course_years, dependent: :delete_all
   belongs_to :course_year_one, class_name: "CourseYear", optional: true
   belongs_to :course_year_two, class_name: "CourseYear", optional: true
   has_many :early_career_teacher_profiles
   has_many :early_career_teachers, through: :early_career_teacher_profiles, source: :user
   has_many :mentor_materials
 
   def course_years
-    [course_year_one, course_year_two].filter(&:present?)
```
asmega, line 12: I've changed this so it returns an ActiveRecord thing so we can chain things if needed. also prevents the 2 initial queries

spec/cypress/integration/CipStoriesAdmin.feature
```diff
@@ -106,4 +106,20 @@ Feature: Admin user interaction with Core Induction Programme
...
-    And the page should be accessible
\ No newline at end of file
+    And the page should be accessible
+
+  Scenario: Can create content-less lessons
+    Given core induction programme with year and module
+    And I am on "core induction programme year" page with id "a4dc302c-ab71-4d7b-a10a-3116a778e8d5"
+
+    When I click on "link" containing "Create CIP Lesson"
+    Then "page heading" should contain "Create lesson"
```
kicferk1, line 116: Let's use 
```
Then I should be on "core induction programme mentor material edit" page
```
step if what we want to check here is the location.

Also, let's add 
```
And the page should be accessible
```

spec/cypress/integration/CipStoriesAdmin.feature
```diff
@@ -106,4 +106,20 @@ Feature: Admin user interaction with Core Induction Programme
...
+
+    When I type "New content-less lesson" into "title input"
+    And I type "30" into "input#course-lesson-completion-time-in-minutes-field"
+    And I type "Some ECT summary" into "#course-lesson-ect-summary-field"
+    And I type "Some mentor summary" into "#course-lesson-mentor-summary-field"
+    And I click on "button" containing "Create lesson"
+
+    Then "page heading" should contain "New content-less lesson"
+    And "page body" should contain "Duration: 30 minutes"
```
kicferk1, line 125:
```
And the page should be accessible
```

spec/requests/core_induction_programme/course_lesson_spec.rb
```diff
@@ -64,6 +65,56 @@
...
+            ect_summary: "new ect summary",
+            completion_time_in_minutes: "10",
+            course_module_id: create(:course_module).id,
+          },
+        }
+
+        expect(response).to redirect_to(lesson_path(CourseLesson.last))
+      end
+
+      context "sad path" do
```
kicferk1, line 103: Maybe let's call it "when given invalid data" or something like that ;)

kicferk1 (commented):
Looks pretty good, some small comments :)

spec/requests/core_induction_programme/course_lesson_spec.rb
asmega: done

spec/cypress/integration/CipStoriesAdmin.feature
asmega: I'll need to remember those in future

spec/cypress/integration/CipStoriesAdmin.feature
asmega: done

app/views/core_induction_programmes/lessons/new.html.erb
```diff
@@ -0,0 +1,31 @@
...
+            width: 4, hint: { text: " For example, 90"},
+            value: @course_lesson.completion_time_in_minutes) %>
+      <%= f.govuk_text_area(
+            :ect_summary,
+            label: { text: 'ECT summary' },
+            rows: 10) %>
+      <%= f.govuk_text_area(
+            :mentor_summary,
+            label: { text: 'Mentor summary' },
+            rows: 10) %>
```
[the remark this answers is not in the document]
asmega: done

app/views/core_induction_programmes/lessons/_course_lesson_fields.html.erb
```diff
@@ -0,0 +1,23 @@
...
+      value: @course_lesson.completion_time_in_minutes) %>
+<%= f.govuk_text_area(
+      :ect_summary,
+      label: { text: 'ECT summary' },
+      rows: 10) %>
+<%= f.govuk_text_area(
+      :mentor_summary,
+      label: { text: 'Mentor summary' },
+      rows: 10) %>
+<%= f.govuk_submit t(f.object.persisted? ? :update : :create, scope: [:helpers, :submit, :course_lesson]) %>
```
asmega, line 23: From what I can tell `govuk_design_system_formbuilder` does not support the default i18n rails looks up so this has been added. if we can fix it upstream we remove this but keep the locale entries

kicferk1: Personally I would have been tempted to have a `Save` text here. 

Or maybe to take the button out of the shared fields and type the version that we wanted in the actual template.

Buuut I don't actually mind it. We will probably spend a while discussing the use of translations on dev retro next week, in the meantime this can go in :)

PR #96, opened 2021-05-10
Start page

callumacrae:
### Context

https://dfedigital.atlassian.net/browse/CPDEL-311

Also added a header fix to this PR which was making the header slightly taller than it should have been when there were no nav items

### Testing

bring back percy </3

### DIFF
--- a/app/views/start/index.html.erb
+++ b/app/views/start/index.html.erb
@@ -1,7 +1,10 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl">Support for early career teachers</h1>
-    <h2 class="govuk-heading-l">Start Page</h2>
-    <%= govuk_link_to "Start", new_user_session_path, button: true %>
+    <h1 class="govuk-heading-xl govuk-!-margin-bottom-6">Support for early career teachers</h1>
+    <p class="govuk-body">This service is for teachers and mentors to access Department for Education-accredited training and support materials.</p>
+    <%= govuk_start_now_button text: "Start now", href: new_user_session_path %>
+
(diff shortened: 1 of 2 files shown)


app/views/start/index.html.erb
```diff
@@ -1,7 +1,10 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl">Support for early career teachers</h1>
-    <h2 class="govuk-heading-l">Start Page</h2>
-    <%= govuk_link_to "Start", new_user_session_path, button: true %>
+    <h1 class="govuk-heading-xl govuk-!-margin-bottom-6">Support for early career teachers</h1>
+    <p class="govuk-body">This service is for teachers and mentors to access Department for Education-accredited training and support materials.</p>
+    <%= govuk_start_now_button text: "Start now", href: new_user_session_path %>
+
+    <h3 class="govuk-heading-m">Before you start</h3>
```
kicferk1, line 7: Are we sure this should be `h3`? We have `h1` but no `h2` :P

kicferk1 (approved):
Looks good, one question about html semantics

app/views/start/index.html.erb
callumacrae: goood point - i just copied the prototype but this will fail the accessibility tests

spec/cypress/integration/accessibility.js
```diff
@@ -1,9 +1,4 @@
 describe("Accessibility", () => {
-  it("Landing page should be accessible", () => {
-    cy.visit("/");
-    cy.checkA11y();
-  });
```
callumacrae, line 5: Was a duplicate test - it's being tested in StartPage.feature too

PR #98, opened 2021-05-11
User research May

kicferk1:
### Context
We want to do UR sessions in Rails app - and the app we get from a PR will serve that purpose.

Some cosmetic changes
Update content snapshot and seed files from staging.

### Guidance to review
We will think about what to keep after UR is done.

### Testing
After UR, when we know which features we wanna keep, we can decide what testing should be done.


kicferk1:
After all was said and done, this PR amounts to:
1. Removing module navigation
2. Improving wording and numbers of self-study materials shown on course module page.

callumacrae (approved):

PR #99, opened 2021-05-12
CPDEL-422: Fix errors to radios not linking to bottom of the page

kicferk1:
### Context
https://dfedigital.atlassian.net/browse/CPDEL-422
There was a bug in progress input error - the error from summary was not taking the user to the actual error message when clicked. 
Also, change copy for https://dfedigital.atlassian.net/browse/CPDEL-433

Use @asmega 's advice to pass a parameter to one of the radio buttons telling it to link the errors to it.

### Guidance to review
Log in as `ucl-early-career-teacher@example.com`, go to any lesson, go to last page part, select nothing and click the submit button. You should see an error. Click on the error in the summary at the top of the page. You should be taken to the radios at the bottom of the page.

### Testing
I am not sure how to test it. I guess in cucumber we could recreate this situation, and click on the error message in the summary, but is there a way in cypress to check that we have scrolled? @callumacrae 

Any other ideas?

### DIFF
--- a/app/models/course_lesson_progress.rb
+++ b/app/models/course_lesson_progress.rb
@@ -7,7 +7,7 @@ class CourseLessonProgress < ApplicationRecord
     complete: "complete",
   }
 
-  validates :progress, presence: { message: "Select if you have finished the session" }
+  validates :progress, presence: { message: "Select if you have finished and understood the session" }
 
   belongs_to :course_lesson
   belongs_to :early_career_teacher_profile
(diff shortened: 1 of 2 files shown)


tonyheadford:
I suppose you _could_ test it by comparing the `href` generated in the error summary with the `id` of the radio button?

callumacrae:
Is there much point in adding a test for it? I don't see how it would happen again 🤷‍♂️

callumacrae (approved):

PR #103, opened 2021-05-13
CPDEL-427: Add a way to send emails to mentors

kicferk1:
### Context
We want to invite mentors as well as ECTs to our service. https://dfedigital.atlassian.net/browse/CPDEL-427

Change the invite job to correctly handle mentors. Add unit tests and readme about it.

### Guidance to review
I will probably need to make sure the job works in the review app by using Rails console.

### Testing
Unit tested the job.


kicferk1:
Looks like it works. I used Rails console to change mentor's email to my own, then I scheduled a task, and I got an email talking about mentoring.
image

lib/tasks/invites.rake
```diff
@@ -2,7 +2,7 @@
 
 namespace :invites do
   desc "Send ECT invites"
```
xswoodsx, line 4: desc should be ECTs and mentors or just participants instead?

xswoodsx (approved):
Looks good other than the picky comment on a desc

PR #104, opened 2021-05-13
[CPDE-383] Sync users on sign in attempt

asmega:
### Context

- User data needs to come from Register & Partner so users can login to view their course

- Updated service so it can now sync existing users from existing Register & Partner api
- Extremely naive implementation where we sync users when a user attempts to login and they do not exist in our database

### Guidance to review

- Not sure if this will work if deployed. ENV vars will probably need to be setup and point to dev Register & Partner
- There will be syncing issues if users change email addresses to existing ones due to validation. Should probably raise a ticket to resolve in the future

### Testing

- update `.env.development` and set `REGISTER_AND_PARTNER_URL` to the dev instance of Register & Partner
- Attempt to sign in as a user that exists on that api endpoint and not your local system
- Should create the user and you should be signed in

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,0 +1,1 @@
+REGISTER_AND_PARTNER_URL=https://api.example.com
--- a/app/models/register_and_partner_api/user.rb
+++ b/app/models/register_and_partner_api/user.rb
@@ -2,8 +2,12 @@
 
 module RegisterAndPartnerApi
   class User < RegisterAndPartnerApi::Resource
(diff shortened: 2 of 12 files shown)


kicferk1:
I checked that `lead-provider@example.com` is pulled from R&P and given a UCL core induction programme. We don't have a user with such an email in our database setup, so it seems the integration works.

xswoodsx (approved):
All looks good to me :)

PR #110, opened 2021-05-19
CPDEL-438: Add authentication to R&P client

kicferk1:
### Context
We added authentication to R&P, so our calls to their api won't work anymore.

Add authentication to our client. If you want the top secret key for R&P dev environment, pm me. 

### Guidance to review
Try logging in as `lead-provider@example.com`. It should work.

### Testing
This is essentially adding one more secret, not sure there is much to test other than the integration actually working in practice.


kicferk1:
After changes from https://github.com/DFE-Digital/early-careers-framework/pull/391 everything should work. I checked that it works with the current version of R&P dev environment db (I logged in as lead provider, yay). 

That db will get wiped and re-created, but the PR I linked to will make sure that we keep the same E&L token that it has now.

xswoodsx (approved):
LGTM :)

PR #113, opened 2021-05-20
Add accessibility tests for mentor materials

callumacrae:
### Context

from the backlog of michal 🧠 

### Testing

yes

### DIFF
--- a/spec/cypress/integration/accessibility.js
+++ b/spec/cypress/integration/accessibility.js
@@ -27,5 +27,16 @@ describe("Accessibility", () => {
         });
       });
     });
+
+    it("Visit all mentor materials to check for accessiblity", () => {
+      cy.app("load_seed");
+      cy.login("mentor");
+      cy.appEval(`MentorMaterial.all.map(&:id)`).then((ids) => {
+        cy.wrap(ids).each((id) => {
+          cy.visit(`/mentor-materials/${id}`);
+          cy.checkA11y();
+        });
+      });
+    });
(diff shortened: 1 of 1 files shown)


spec/cypress/integration/accessibility.js
```diff
@@ -27,5 +27,16 @@ describe("Accessibility", () => {
         });
       });
     });
+
+    it("Visit all mentor materials to check for accessiblity", () => {
+      cy.app("load_seed");
+      cy.login("mentor");
```
kicferk1, line 33: We should probably log in as admin, mentors won't (in principle) see all of them.

That said, do we also need to login as admin in the lesson part test, now that we correctly restrict access to them?

spec/cypress/integration/accessibility.js
```diff
@@ -17,9 +17,10 @@ describe("Accessibility", () => {
 
   // This test should only be ran locally due to the length of time taken to complete.
   // To include it add '--env tags=checkCourseLessonsAccessibility' to the yarn cypress:open cmd.
-  if (Cypress.env("tags")?.includes("checkCourseLessonsAccessibility")) {
+  if (!Cypress.env("tags")?.includes("checkCourseLessonsAccessibility")) {
```
kicferk1, line 20: You sure about that? 😂

kicferk1 (approved):
The negation in when we run the tests should probably be reverted, but other than that looks good to me.

PR #116, opened 2021-05-20
[CPDEL-454] Add reg and partner field to user

xswoodsx:
### Context
Users in register and partner can change their email address. However this is what we're currently using to check if they exist on our service. R&P send us the users id so we can store this on our side, avoiding us creating duplicate users. [Jira ticket](https://dfedigital.atlassian.net/browse/CPDEL-454)

Migration to add `register_and_partner_id` to Users table
Changes sync_users method to find user by their register_and_partner_id

### Guidance to review
Having the same user in the users.json file but with a changed email feels like cheating but i couldn't find a simpler way of testing the email updated. 

### Testing
Additional test for making sure the user isn't created twice and their email is updated. As above with the cheating. 

You could try it manually by following [these steps](https://github.com/DFE-Digital/ecf-engage-and-learn/blob/develop/documentation/register_and_partner_api_setup.md#developing-against-a-local-rp) to get the API running. Once you've done that, login on engage and learn with `lead-provider@example.com`. Change the `lead-provider@example.com` email address in your local R&P db, try to login with this new email on engage and learn and it should still work. The new email should also be stored on the engage and learn side.

### DIFF
--- a/spec/services/register_and_partner_api/sync_users_spec.rb
+++ b/spec/services/register_and_partner_api/sync_users_spec.rb
@@ -25,11 +25,13 @@ RSpec.describe RegisterAndPartnerApi::SyncUsers do
     end
 
     it "updates an existing user that has changed their email address" do
+      create(:user, register_and_partner_id: "65abf54c-c7c2-490b-9bcd-bbb4e0aab934", email: "lead-provider-1@example.com")
+
       described_class.perform
 
       record = User.find_by(register_and_partner_id: "65abf54c-c7c2-490b-9bcd-bbb4e0aab934")
       expect(User.count).to eql(2)
-      expect(record.email).to eql("lead-provider-1@example.com")
+      expect(record.email).to eql("lead-provider@example.com")
     end
   end
 end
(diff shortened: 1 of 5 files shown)


spec/services/register_and_partner_api/sync_users_spec.rb
```diff
@@ -23,5 +23,13 @@
...
       expect(record.full_name).to eql("Induction Tutor")
     end
+
+    it "updates an existing user that has changed their email address" do
+      described_class.perform
+
+      record = User.find_by(register_and_partner_id: "65abf54c-c7c2-490b-9bcd-bbb4e0aab934")
+      expect(User.count).to eql(2)
+      expect(record.email).to eql("lead-provider-1@example.com")
+    end
```
kicferk1, line 33: An easier way would be to set up a user in the test, before you call `perform`, and give it a different email but the same id. 

Then you can avoid the duplicated user in our `users.json` file :)

kicferk1 (approved):

PR #117, opened 2021-05-24
Update user sync

xswoodsx:
### Context
We can now receive the type of user and their cip from the R&P api. We want to use this to create the right profile and assign their cip to them on E&L.

Updated user sync method that creates a users profile based on the user_type sent and assigns them a cip based on the cip_type. 

### Guidance to review
We're currently only doing this for ECTs as we only receive a type of `early_career_teacher` and `other`. The other type will create a mentor profile. 

See [R&P PR](https://github.com/DFE-Digital/early-careers-framework/pull/400) for further info on the changes made on their side to see how what data is being sent and how it's selected. An enum is used for sending CIP_TYPE and USER_TYPE

### Testing
A couple changes to the sync_users_spec and an additional test to check user type.

### DIFF
--- a/app/services/register_and_partner_api/sync_users.rb
+++ b/app/services/register_and_partner_api/sync_users.rb
@@ -2,6 +2,19 @@
 
 module RegisterAndPartnerApi
   class SyncUsers
+    USER_TYPES = {
+      ect: "early_career_teacher",
+      mentor: "mentor",
+    }.freeze
+
+    CIP_TYPES = {
+      ambition: "Ambition Institute",
+      edt: "Education Development Trust",
+      teach_first: "Teach First",
+      ucl: "UCL",
+      none: "none",
+    }.freeze
+
     def self.perform
       # TODO: Add the last time we synced users, to avoid getting too many of them back
       # TODO: Add pagination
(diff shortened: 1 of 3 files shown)


app/services/register_and_partner_api/sync_users.rb
```diff
@@ -38,9 +39,40 @@ def self.user_attributes_from(user_from_api)
...
+      when "edt"
+        "Education Development Trust"
+      when "teach_first"
+        "Teach First"
+      when "ucl"
+        "UCL"
+      end
+    end
+
+    def self.set_profile_type(user, user_type)
```
kicferk1, line 66: `set_profile_type` is not a great name - the method actually returns us a profile based on user type.

`find_or_initialize_profile` would be a bit better :)

app/services/register_and_partner_api/sync_users.rb
```diff
@@ -38,9 +39,40 @@ def self.user_attributes_from(user_from_api)
...
+
+    def self.set_core_induction_programme(attributes, profile)
+      profile.core_induction_programme = CoreInductionProgramme.find_by(
+        name: determine_core_induction_programme(attributes[:cip]),
+      )
+    end
+
+    def self.determine_core_induction_programme(cip_name)
+      case cip_name
+      when "ambition"
```
kicferk1, line 55: Let's copy-paste the constant `CIP_TYPES` from R&P repo to be a bit more certain that we have the same options as what the API returns.

app/services/register_and_partner_api/sync_users.rb
```diff
@@ -38,9 +39,40 @@ def self.user_attributes_from(user_from_api)
...
+      end
+    end
+
+    def self.set_profile_type(user, user_type)
+      case user_type
+      when "early_career_teacher"
+        ::EarlyCareerTeacherProfile.find_or_initialize_by(user: user)
+      when "other"
+        ::MentorProfile.find_or_initialize_by(user: user)
+      end
```
kicferk1, line 72: I would say when `other`, we shouldn't do anything at all. 

Maybe add `mentor` type just on our side, and make a mentor profile for them, but for `other` user type we should do nothing, not even create a user (I think).

xswoodsx: It doesn't seem like not having something assigned to profile before calling save on user. 

If we call user.save! before assigning a profile it works. Unless there
The below works when we don't create a mentor

`user.save!
profile = find_or_initialize_profile(user, attributes[:user_type])
set_core_induction_programme(attributes, profile)
profile.save!`

kicferk1 (approved):
I would add a test showing what happens to the "other" type of user. Other than that looks good to me :)

PR #118, opened 2021-05-24
Changes to year page for mentors

callumacrae:
https://dfedigital.atlassian.net/browse/CPDEL-469

- Display mentor description when specified
- Add mentor-specific titles
- Make description optional

### Testing

yes


app/views/core_induction_programmes/years/_module_list.html.erb
```diff
@@ -7,7 +7,16 @@
         </h3>
 
         <div class="gem-c-govspeak govuk-govspeak">
-          <%= course_module.ect_summary_to_html.html_safe %>
+          <% if current_user.admin? %>
+            <h4 class="govuk-heading-s">Mentor summary</h4>
+            <%= course_module.mentor_summary_to_html.html_safe.presence || "<p>No mentor summary</p>".html_safe %>
```
callumacrae, line 12: Not sure about this, is there a better way to do it?

kicferk1:
```
<% if course_module.mentor_summary %>
  <%= course_module. mentor_summary_to_html.html_safe %>
<% else %>
  <p>No mentor summary</p>
<% end %>
```
I guess

kicferk1: Or go to `mentor_summary_to_html` method and add that logic for empty value being `<p>No mentor summary</p>` in there.

app/models/course_year.rb
```diff
@@ -48,6 +49,12 @@ def core_induction_programme
     core_induction_programme_one || core_induction_programme_two
   end
 
+  def title_for(user)
+    return title if mentor_title.blank?
+
+    user.mentor? ? mentor_title : title
+  end
```
kicferk1, line 56: We should unit test it in course yer model spec.

PR #119, opened 2021-05-25
Show and Tell - do not merge

kicferk1:


kicferk1:
Show and tell is over, and so usefulness of this app has ended.

PR #121, opened 2021-05-25
Add guidance for mentors

callumacrae:
### Context

https://dfedigital.atlassian.net/browse/CPDEL-408

### Testing

yes


app/controllers/training_and_support_controller.rb
```diff
@@ -9,5 +9,7 @@ def show
     if @provider.blank?
       raise Pundit::NotAuthorizedError, "This page is only available to Early Career Teachers"
     end
+
+    @provider_name_key = @provider.name.downcase.gsub(/ /, "_")
```
kicferk1, line 13: The error is a bit wrong - maybe change `Early Career Teachers` to `CIP participants`

kicferk1 (approved):
Tiny error wording nit-pick, which probably doesn't matter much.

PR #122, opened 2021-05-26
Mentor dashboard

callumacrae:
### Context

https://dfedigital.atlassian.net/browse/CPDEL-418

### Testing

yes

### DIFF
--- a/app/views/dashboard/_dashboard_ect.html.erb
+++ b/app/views/dashboard/_dashboard_ect.html.erb
@@ -1,0 +1,20 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <div class="blue-header">
+      <h1 class="govuk-heading-xl blue-header__title">Welcome <%= current_user.name %></h1>
+      <strong><%= govuk_link_to "Edit name", edit_preferred_name_path, class: "blue-header__text", "data-test": "edit-name" %></strong>
+    </div>
+
+    <% if current_user.early_career_teacher_profile.core_induction_programme %>
+      <%= govuk_link_to "Go to your module materials",
+        cip_path(current_user.early_career_teacher_profile.core_induction_programme),
+        button: true %>
+    <% end %>
+
+    <h2 class="govuk-heading-l govuk-!-margin-top-4">Related content</h2>
+
+    <p><%= govuk_link_to "Information about training and support during your induction", training_and_support_path %></p>
+    <p><%= govuk_link_to "How the teacher standards relate to the ECF", "#" %></p>
+    <p><%= govuk_link_to "Your mentor's role and responsibilities", "#" %></p>
+  </div>
+</div>
--- a/app/views/dashboard/_dashboard_ect.html.erb
+++ b/app/views/dashboard/_dashboard_ect.html.erb
@@ -5,7 +5,7 @@
       <strong><%= govuk_link_to "Edit name", edit_preferred_name_path, class: "blue-header__text", "data-test": "edit-name" %></strong>
     </div>
 
-    <% if current_user.early_career_teacher_profile.core_induction_programme %>
+    <% if current_user.early_career_teacher_profile&.core_induction_programme %>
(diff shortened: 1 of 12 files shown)


app/views/dashboard/_dashboard_mentor.html.erb
```diff
@@ -0,0 +1,40 @@
...
+    <div class="govuk-width-container">
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+          <h1 class="govuk-heading-xl blue-header__title">Plan compelling feedback and coaching sessions</h1>
+          <p class="blue-header__text govuk-body-l">Using the materials available in this service, you can support your school’s early career teachers develop the skills they need to thrive.</p>
+
+          <% if current_user.mentor? && current_user.mentor_profile.core_induction_programme %>
+            <%= render GovukComponent::StartNowButton.new text: "Start now",
+              href: cip_path(current_user.mentor_profile.core_induction_programme) %>
+          <% end %>
```
kicferk1, line 23: The user will always be a mentor here, right?

app/views/dashboard/show.html.erb
```diff
@@ -1,27 +1,5 @@
...
-    <p><%= govuk_link_to "How the teacher standards relate to the ECF", "#" %></p>
-    <p><%= govuk_link_to "Your mentor's role and responsibilities", "#" %></p>
-
-  </div>
-</div>
+<% if current_user.mentor? %>
+  <%= render "dashboard_mentor" %>
+<% else %>
+  <%= render "dashboard_ect" %>
+<% end %>
```
kicferk1, line 5: Can we have a check that user is an ect before we show them ect dashboard, and maybe have a `default_dashboard` shown otherwise?

kicferk1: Also, just to double check, are we absolutely sure that we want to make dashboards look very different for different users?

app/views/dashboard/_dashboard_ect.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    <div class="blue-header">
+      <h1 class="govuk-heading-xl blue-header__title">Welcome <%= current_user.name %></h1>
+      <strong><%= govuk_link_to "Edit name", edit_preferred_name_path, class: "blue-header__text", "data-test": "edit-name" %></strong>
+    </div>
+
+    <% if current_user.early_career_teacher_profile.core_induction_programme %>
+      <%= govuk_link_to "Go to your module materials",
+        cip_path(current_user.early_career_teacher_profile.core_induction_programme),
+        button: true %>
+    <% end %>
```
kicferk1, line 12: Pretty sure it will break for admins

kicferk1: Cause they don't have a ect profile, and so nil has no method 😢 

If you add a default dashboard and only show this one for ects it should be good.

app/views/dashboard/show.html.erb
callumacrae: i'm kind of hoping the ECT dashboard changes to match the mentor dashboard, but yes, they're currently very different in the designs

I don't think it's really worth the effort of making a default dashboard unless there's something specific that would be gained from it - I don't think there's any problem with showing them the ECT dashboard

app/views/dashboard/show.html.erb
```diff
@@ -1,27 +1,5 @@
...
-    </div>
-
-    <% if current_user.early_career_teacher? %>
-      <%= govuk_link_to "Go to your module materials",
-        cip_path(current_user.early_career_teacher_profile.core_induction_programme),
-        button: true %>
-    <% end %>
-
-    <% if current_user.mentor? && current_user.mentor_profile.core_induction_programme %>
-      <%= govuk_link_to "Go to your mentee's module materials",
```
kicferk1, line 15: We have a cypress test that uses that. I think the test needs to be updated to use the copy from the start button :P

kicferk1 (approved):

PR #128, opened 2021-05-28
Add sign-in by magic links for admin users

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDEL-453

We want to hide admin users behind proper authentication. 
Since we use staging environment as our source of truth for content, it is really convenient to keep a dummy admin without that requirement in there, but I added a second admin for whom we can test that email sending works.

Our `passwordless_authenticatable` was serving largely as email validator. I decided to move it to the controller.

I copied most of this stuff from R&P, but there are some differences. For one, I think it's much more convenient to specify the test users in the strategy, rather than controller. The other is the obvious one, ects and mentors are still logged in without authentication.

One more change I copied from R&P is to have signed out page in the sessions controller, with all other user sign-in actions.

There are some changes in cypress - cause the login flow for admins is fundamentally different to the one for other users. Maybe I could check if traits include admin to avoid adding more steps. I'll let you judge whether it's a better idea.

### Review Checks

2. Try logging in as `ecf-engage-and-learn@mailinator.com`
3. Go to https://www.mailinator.com/v4/public/inboxes.jsp?to=ecf-engage-and-learn
4. There should be a sign-in email there, and when you click on sign-in link you should be able to log in.


app/controllers/users/sessions_controller.rb
```diff
@@ -1,13 +1,89 @@
...
+  end
+
+  def sign_in_with_token
+    @user.update!(login_token: nil, login_token_valid_until: 1.year.ago)
+    sign_in(@user, scope: :user)
+    redirect_to after_sign_in_path_for(@user)
+  end
+
+  def redirect_from_magic_link
+    @login_token = params[:login_token] if params[:login_token].present?
```
xswoodsx, line 27: Is it finding any existing login_token from the session cookies?

app/mailers/user_mailer.rb
```diff
@@ -3,6 +3,7 @@
 class UserMailer < ApplicationMailer
   MENTOR_WELCOME_TEMPLATE = "c0dc7dae-76e3-4346-8c2b-0d38aaf94a54"
   ECT_WELCOME_TEMPLATE = "652fea63-1344-4608-a957-6046dc27120b"
+  SIGN_IN_EMAIL_TEMPLATE = "a3219fe5-e320-48ee-a386-7a244785785c"
```
xswoodsx, line 6: What's the string in this relate to? A template you'd expect to see Dear X etc. Is a like a reference to a template in notify?

config/environments/development.rb
```diff
@@ -19,7 +19,7 @@
   config.domain = ENV["DOMAIN"]
   config.gias_api_root = ENV["GIAS_API_ROOT"]
   # Don't care if the mailer can't send.
-  config.action_mailer.raise_delivery_errors = false
+  config.action_mailer.raise_delivery_errors = true
```
xswoodsx, line 22: This will now tell us when an email hasn't been sent because of a "dodgy" email address or we can a failed to send response?

config/environments/test.rb
```diff
@@ -46,7 +46,7 @@
   config.action_mailer.default_options = {
     from: "mail@example.com",
   }
-  config.domain = "localhost:3000"
+  config.domain = "www.example.com"
```
xswoodsx, line 49: What's the significance of changing this?

spec/cypress/integration/CipLessons.feature
```diff
@@ -2,7 +2,7 @@ Feature: Core Induction Programme lessons
   Users should be able to view and sometimes edit cip lessons.
 
   Scenario: Admins can edit lessons
-    Given I am logged in as "admin"
+    Given I am admin and logged in
```
xswoodsx, line 5: I hadn't seen how the Given statements were made. I half assumed it was Percy wizadry but is it matching the "I am admin and logged in" and then running the function loginAdmin() written in Cypress? So Given takes a string and function as its parameters?

app/controllers/users/sessions_controller.rb
kicferk1: Nah, it's finding it from the url. The url we generate will look something like 
```
https://ecf-engage-and-learn-review-pr-128.london.cloudapps.digital/users/confirm-sign-in?login_token=81ed78c85706afc0546b
```
Note the last part: `?login_token=81ed78c85706afc0546b`. `?` in it means that the main part of the url is over, and that we are starting to include url parameters. `login_token` is the name of the parameter, and `81ed78c85706afc0546b` is the value.

And Rails helpfully makes this available in the controller `params`.

kicferk1: I guess if it was not in the url, but you edited your session cookie to include `login_token` key, it might be able to find it, maybe. Although probably not inside `params`.

app/mailers/user_mailer.rb
kicferk1: It relates to https://www.notifications.service.gov.uk/services/ea6cc52b-8cba-4568-9293-14d7049bc753/templates/a3219fe5-e320-48ee-a386-7a244785785c

config/environments/development.rb
kicferk1: This will tell us if Notify refuses to send our email. This would happen if we tried to send something with a test api key, to someone outside of our tam and whitelist.

config/environments/test.rb
kicferk1: It makes links in tests look more obviously 'test' like. Localhost links look much more like they belong in develop environment.

spec/cypress/integration/CipLessons.feature
kicferk1: Let's chat about it in person on Tuesday :)

The short gist of it is that we define these steps in `spec/cypress/support/step_definitions/factory-bot.js`.

app/controllers/users/sessions_controller.rb
xswoodsx: I think i was thinking about how a session persists when you leave the site and then come back but are still logged in. I thought this could be doing that as well as what you've mentioned above.  But yeah thinking about it, that wouldn't be happening here because it wouldn't enter from the redirect from magic link.

callumacrae:
Instead of combining the authentication strategies into one strategy, did you work out if there's a way to have a different strategy for different user types?

spec/cypress/support/step_definitions/factory-bot.js
```diff
@@ -51,6 +51,24 @@ const login = (traits, args) => {
...
+    .as("userData")
+    .then(([user]) => {
+      cy.visit(`/users/confirm-sign-in?login_token=${user.login_token}`);
+    });
+
+  cy.get('[action="/users/sign-in-with-token"] [name="commit"]').click();
+};
+
+Given("I am admin and logged in", () => loginAdmin());
+Given("I am admin and logged in with {}", (args) => loginAdmin(args));
```
callumacrae, line 71: I'd prefer it if these were just a special case in the existing step definitions (and that would be backwards compatible, too)

kicferk1:
> Instead of combining the authentication strategies into one strategy, did you work out if there's a way to have a different strategy for different user types?

I played around with it for a while, it didn't want to work how I wanted it (making it a single controller and a single model seems to be hard), and I decided to not worry about it until we really need a different strategy :/

spec/cypress/support/step_definitions/factory-bot.js
kicferk1: I made admin a special case in our definitions, looks a bit better to me :)

lib/devise/strategies/passwordless_authenticatable.rb
```diff
@@ -0,0 +1,61 @@
...
+          end
+        end
+      end
+
+    private
+
+      def user_requires_magic_link(user)
+        return false unless user&.admin?
+
+        in_test_environment = Rails.env.development? || Rails.env.deployed_development? || Rails.env.staging?
```
callumacrae, line 52: I particularly like the way that `in_test_environment` checks for everything _but_ the `test` environment 😝

callumacrae (approved):

PR #130, opened 2021-06-01
CPDEL-518: Sync users periodically

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDEL-518

Going commit-by-commit should make things more understandable. I decided to add a worker app to staging, to see how it will perform. Copied that one from R&P. 

### Review Checks

2. So long as you clicked 5 minutes or later from the app being created, you should be able to log in with `rp-ect-ambition@example.com` - a user that we can only get from their API.

### DIFF
--- a/terraform/app/modules/paas/main.tf
+++ b/terraform/app/modules/paas/main.tf
@@ -11,6 +11,7 @@ resource cloudfoundry_app web_app {
   docker_image = var.app_docker_image
   health_check_type = "http"
   health_check_http_endpoint = "/check"
+  health_check_timeout = 60
   instances = var.web_app_instances
   memory = var.web_app_memory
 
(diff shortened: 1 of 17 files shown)


terraform/app/modules/paas/main.tf
```diff
@@ -31,7 +32,29 @@ resource cloudfoundry_app web_app {
 }
 
 resource cloudfoundry_route web_app_route {
-  domain   = data.cloudfoundry_domain.cloudapps_digital.id
-  space    = data.cloudfoundry_space.space.id
+  domain = data.cloudfoundry_domain.cloudapps_digital.id
+  space = data.cloudfoundry_space.space.id
```
ethanmills, line 36: No change here, but I think we should start running `terraform fmt` to make this consistent

ethanmills (approved):
Looks good to me. I think E&L is quickly going to become the biggest source of R&P traffic 😂

kicferk1:
> Looks good to me. I think E&L is quickly going to become the biggest source of R&P traffic 😂

Once every 5 minutes, making it 288 requests per day xD

PR #131, opened 2021-06-03
Send invites on user sync

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDEL-490

Commit by commit should be pretty straightforward.

I put the logic for whether we want to create an email or not in the sync service. I guess it could live in an after create hook on a user, but at that time we might not yet know whether the user is a mentor or an ect.

I put the logic for whether to send an email or not in the email model.

When we get ECTs, we will create invite emails for them, but we won't send them. At some point we will probably need a job for sending unsent emails.

I added email sending for mentors too, because we want it, and it was easier to add it here than to split it into its own PR. 

### Review Checks

1. You can't really. If you want to check that mentors receive an email, we would need to add a mentor that is whitelisted on our notify instance to R&P dev environment, wait 5 minutes for user sync to happen, and then the mentor email should be sent.

### DIFF
--- a/app/models/invite_email_ect.rb
+++ b/app/models/invite_email_ect.rb
@@ -1,7 +1,15 @@
 # frozen_string_literal: true
 
 class InviteEmailEct < TrackedEmail
+  INVITES_SENT_FROM = Date.new(2021, 8, 20).beginning_of_day
+
   def mail_to_send
     UserMailer.ect_welcome_email(user)
+  end
+
+  def send!
+    if Time.zone.now >= INVITES_SENT_FROM
+      super
+    end
   end
 end
--- a/app/services/invite_participants.rb
+++ b/app/services/invite_participants.rb
@@ -8,9 +8,6 @@ class InviteParticipants
     emails.each do |email|
       user = User.find_by(email: email)
       email = create_invite_email_for_user(user)
-
-      # if user is ect and before certain time, don't send
-      # otherwise send
       email.send!
     rescue StandardError
       logger.info "Error emailing user, id: #{user&.id} ... skipping"
@@ -19,9 +16,9 @@ class InviteParticipants
 
   def self.create_invite_email_for_user(user)
     if user.mentor?
-      InviteEmailMentor.create!(user: user)
+      InviteEmailMentor.find_or_create_by!(user: user, sent_at: nil)
     elsif user.early_career_teacher?
-      InviteEmailEct.create!(user: user)
+      InviteEmailEct.find_or_create_by!(user: user, sent_at: nil)
     end
   end
 
--- a/app/services/invite_participants.rb
+++ b/app/services/invite_participants.rb
@@ -1,22 +1,23 @@
 # frozen_string_literal: true
 
 class InviteParticipants
-  def run(emails)
+  def self.run(emails)
     logger.info "Emailing Participants"
 
     user = nil
     emails.each do |email|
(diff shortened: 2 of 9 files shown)


app/services/register_and_partner_api/sync_users.rb
```diff
@@ -54,41 +54,39 @@ def self.user_attributes_from(user_from_api)
...
-        CIP_TYPES[:teach_first]
-      when "ucl"
-        CIP_TYPES[:ucl]
-      end
-    end
-
     def self.find_or_create_user(attributes, profile_class)
       user = ::User.find_or_initialize_by(register_and_partner_id: attributes[:register_and_partner_id])
-      profile = profile_class.find_or_initialize_by(user: user)
+      needs_inviting = !user.persisted?
```
xswoodsx, line 59: negation sozzle my head at times. what's !user checking with this persisted?

app/services/invite_participants.rb
```diff
@@ -1,30 +1,33 @@
 # frozen_string_literal: true
 
 class InviteParticipants
-  def run(emails)
+  def self.run(emails)
     logger.info "Emailing Participants"
 
     user = nil
```
xswoodsx, line 7: It was already in but why is user assigned nil and then assigned again on line 9?

lib/tasks/invites.rake
```diff
@@ -3,6 +3,6 @@
 namespace :invites do
   desc "Send invites"
   task :send_invites, [:emails] => :environment do |_task, args|
-    InviteParticipants.new.run(args.emails.split)
```
xswoodsx, line 6: What's the significance of removing the new?

app/models/invite_email_ect.rb
```diff
@@ -1,7 +1,15 @@
 # frozen_string_literal: true
 
 class InviteEmailEct < TrackedEmail
+  INVITES_SENT_FROM = Date.new(2021, 8, 20).beginning_of_day
```
xswoodsx, line 4: We've added this date because that's when we're planning to start sending emails from?

app/services/invite_participants.rb
```diff
@@ -1,30 +1,33 @@
...
     emails.each do |email|
       user = User.find_by(email: email)
```
xswoodsx: It was already in but why is user assigned nil and then assigned again on line 9?

kicferk1: So that if an exception happens the `user` is defined (and `nil`). If it's not, then the code trying to log the exception will throw the exception itself :D

app/models/invite_email_ect.rb
kicferk1: More or less, the exact date is still in the works, I think, but yeah.

lib/tasks/invites.rake
kicferk1: You can very easily stub and test class methods. Stubbing and testing instance methods did not seem nearly as easy when I tried it, so I decided to switch to using class methods instead.

app/services/register_and_partner_api/sync_users.rb
kicferk1: `user.persisted?` checks whether the user was already saved in the database. `!user.persisted?` is a negation of that, so it checks whether the user was never saved - so whether they are a new user.

PR #144, opened 2021-06-10
Add prod environment

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDEL-550

We want a live environment, even if it is not fully live yet.
When this gets merged I will do a staging release, and then a prod release. And it hopefully _just works_.

I copied deployment setup from a recent version of R&P.

The notify key is not live, and the R&P token is not real - those will come later. Monitoring will come later. If something fancy but not required to deploy an app is missing here, well, it will probably come later. Do mention it though.

### Review Checks


.gitignore
```diff
@@ -54,3 +54,5 @@ tokens
 /spec/cypress/screenshots
 
 /config/credentials/staging.key
+
+/config/credentials/production.key
```
ethanmills, line 58: This is fine, but you might consider changing to `/config/credentials/*.key` to make it even more foolproof

terraform/workspace-variables/production.tfvars
```diff
@@ -0,0 +1,15 @@
...
+paas_api_url = "https://api.london.cloud.service.gov.uk"
+paas_space_name = "early-careers-framework-prod"
+paas_postgres_service_plan = "small-11"
+paas_app_start_timeout = "180"
+paas_app_stopped = false
+paas_web_app_deployment_strategy = "blue-green-v2"
+paas_web_app_instances = 1
+paas_web_app_memory = 8192
+paas_worker_app_instances = 1
+paas_worker_app_start_command = "bundle exec rake jobs:work"
```
ethanmills, line 15: I assume this is deliberately different?

terraform/workspace-variables/production_app_env.yml
```diff
@@ -0,0 +1,10 @@
+---
+RACK_ENV: production
+RAILS_ENV: production
+NODE_ENV: production
+RAILS_LOG_TO_STDOUT: true
+RAILS_MAX_THREADS: 5
+RAILS_SERVE_STATIC_FILES: true
+SERVICE_TYPE: web
+DATABASE_URL: postgres://postgres@localhost:5432
+REGISTER_AND_PARTNER_URL: "https://manage-training-for-early-career-teachers.education.gov.uk/"
```
ethanmills, line 10: Does this mean we're going to get polled with a bad API token? 😂 That's fine, I'll just know it ignore it

ethanmills (approved):
Looks good, one minor suggestion (optional)

terraform/workspace-variables/production_app_env.yml
kicferk1: Yeah, I expect there will be a 401 or 403 every 5 mins :P Sorry about that, want to test it out in staging first, but I don't want our prod to have your staging ECTs and mentors :D

jamesrose (approved):
LGTM :shipit:

terraform/workspace-variables/production.tfvars
kicferk1: Yeah, I haven't played with mailer pools yet, will change it when I'm sure we want it :)

ethanmills: Just for future reference, if you ever call `deliver_later` that should end up in the mailers pool. It will still get send this way, I just didn't want it being blocked by a long running task on ours

.gitignore
kicferk1: like 10 lines higher we already have it 😂 
I just never thought about it cause creating credentials adds the one-by-one ignores :P

blackrat: Added by ECF-TECH-33

PR #146, opened 2021-06-11
Add real token to production credentials

kicferk1:
## Ticket and context

Tested it on staging envs, it seems to work.


ethanmills (approved):
Thrilling read, 10/10

PR #147, opened 2021-06-14
Add logstash

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDEL-550

Copied pretty much everything from R&P.
I intend to see if I get any logs from E&L in the existing logstash setup, if yes it is probably working.

### DIFF
--- a/config/environments/deployed_development.rb
+++ b/config/environments/deployed_development.rb
@@ -105,7 +105,7 @@ Rails.application.configure do
 
   # Include params in logs: https://github.com/roidrage/lograge#what-it-doesnt-do
   config.lograge.custom_options = lambda do |event|
-    exceptions = %w[controller action format id]
+    exceptions = %w[controller action format id login_token]
     {
       params: event.payload[:params].except(*exceptions),
       exception: event.payload[:exception], # ["ExceptionClass", "the message"]
--- a/config/environments/deployed_development.rb
+++ b/config/environments/deployed_development.rb
@@ -89,10 +89,27 @@ Rails.application.configure do
   # require 'syslog/logger'
   # config.logger = ActiveSupport::TaggedLogging.new(Syslog::Logger.new 'app-name')
 
-  if ENV["RAILS_LOG_TO_STDOUT"].present?
-    logger           = ActiveSupport::Logger.new(STDOUT)
-    logger.formatter = config.log_formatter
-    config.logger    = ActiveSupport::TaggedLogging.new(logger)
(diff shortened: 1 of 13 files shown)


kicferk1:
I got some logs through (you can see them under [this long link](https://kibana.logit.io/s/51151c79-e17d-4407-be3a-01960f3f65ff/app/kibana#/discover?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2021-06-14T12:55:30.000Z',to:'2021-06-14T12:56:00.000Z'))&_a=(columns:!(_source),filters:!(),index:'8ac115c0-aac1-11e8-88ea-0383c11b333a',interval:auto,query:(language:kuery,query:'cf.app:%20%22ecf-engage-and-learn-review-pr-147%22'),sort:!()))). So I guess it works.

config/environments/deployed_development.rb
```diff
@@ -89,10 +89,27 @@
...
+  # Use Lograge for cleaner logging
+  config.lograge.enabled = true
+  config.lograge.base_controller_class = ["ActionController::API", "ActionController::Base"]
+  config.lograge.formatter = Lograge::Formatters::Logstash.new
+  config.lograge.ignore_actions = ["ApplicationController#check"]
+  config.lograge.logger = ActiveSupport::Logger.new($stdout)
+
+  # Include params in logs: https://github.com/roidrage/lograge#what-it-doesnt-do
+  config.lograge.custom_options = lambda do |event|
+    exceptions = %w[controller action format id]
```
jamesrose, line 108: Are there any secure tokens or passwords that we might want to exclude here?

kicferk1: Hmm, all I can think of is `login_token`, which I guess we should hide - it isn't very long lived, but if someone failed to log in with a valid token due to some exception, and someone had the logs, and they tried to log in with the same token, they might be able to get in.

Other than that, I don't think we have anything relating to security in params.

ethanmills (approved):

config/environments/deployed_development.rb
kicferk1: Actually, we add 
```
Rails.application.config.filter_parameters += %i[
  passw secret token _key crypt salt certificate otp ssn
]
```
in one of our initialisers, and https://guides.rubyonrails.org/configuring.html tells me it works "by partial matching regular expression", so anything with a `token` should be hidden already.

PR #151, opened 2021-06-15
Cpdel 549 stop fip user email and login

xswoodsx:
## Ticket and context

Ticket: [549](https://dfedigital.atlassian.net/jira/software/projects/CPDEL/boards/89?assignee=5f3543bb3e9e2e004db3189d&selectedIssue=CPDEL-549)

The addition of induction_programme_choice to has been made in [PR-518](https://github.com/DFE-Digital/early-careers-framework/pull/518) on Register and Partner. 

We only store the induction programme choice for synced users. Do we want to set it default to core for all users?

### Review Checks

2. Try to login as a cip user, `rp-mentor-ambition@example.com`, you should be able to login.
3. Try to login as a fip user, `rp-ect-ucl@example.com` or `rp-mentor-ucl@example.com and you should see an error page. 

You should see 

Screenshot 2021-06-16 at 10 47 44

### DIFF
--- a/app/models/early_career_teacher_profile.rb
+++ b/app/models/early_career_teacher_profile.rb
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 
 class EarlyCareerTeacherProfile < ApplicationRecord
+  enum induction_programme_choice: {
+    core_induction_programme: "core_induction_programme",
+    full_induction_programme: "full_induction_programme",
+    not_yet_known: "not_yet_known",
+  }
+
   belongs_to :user
   belongs_to :core_induction_programme, optional: true
   belongs_to :cohort, optional: true
--- a/app/models/early_career_teacher_profile.rb
+++ b/app/models/early_career_teacher_profile.rb
@@ -4,6 +4,8 @@ class EarlyCareerTeacherProfile < ApplicationRecord
   enum induction_programme_choice: {
     core_induction_programme: "core_induction_programme",
     full_induction_programme: "full_induction_programme",
+    design_our_own: "design_our_own",
+    no_early_career_teachers: "no_early_career_teachers",
     not_yet_known: "not_yet_known",
   }
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -37,13 +37,13 @@ class User < ApplicationRecord
   end
 
   def is_on_full_induction_programme?
-    early_career_teacher_profile&.induction_programme_choice == "full_induction_programme" ||
-      mentor_profile&.induction_programme_choice == "full_induction_programme"
+    early_career_teacher_profile&.full_induction_programme? ||
+      mentor_profile&.full_induction_programme?
   end
 
   def is_on_core_induction_programme?
-    early_career_teacher_profile&.induction_programme_choice == "core_induction_programme" ||
-      mentor_profile&.induction_programme_choice == "core_induction_programme"
+    early_career_teacher_profile&.core_induction_programme? ||
+      mentor_profile&.core_induction_programme?
   end
 
   def course_years
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -37,10 +37,7 @@ class User < ApplicationRecord
   end
 
   def is_not_on_core_induction_programme?
-    is_on_full_induction_programme? ||
-      is_on_designed_programme? ||
-      programme_has_no_early_career_teachers? ||
-      programme_not_yet_known?
+    !is_on_core_induction_programme? && (mentor? || early_career_teacher?)
   end
 
   def is_on_core_induction_programme?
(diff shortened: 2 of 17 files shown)


kicferk1:
For product review, is there an R&P user that is doing FIP?

app/models/user.rb
```diff
@@ -36,6 +36,16 @@ def core_induction_programme
     mentor_profile.core_induction_programme if mentor?
   end
 
+  def is_on_full_induction_programme?
+    early_career_teacher_profile&.induction_programme_choice == "full_induction_programme" ||
+      mentor_profile&.induction_programme_choice == "full_induction_programme"
+  end
```
kicferk1, line 42: If you make it an enum on the profiles, you will be able to write
```
early_career_teacher_profile&.full_induction_programme? || mentor_profile&.full_induction_programme?
```
It will also throw an error when it gets a value outside of the enum values, which is a great safeguard against R&P changing their API in the future and silently messing up our database. 

Something like `CourseLessonProgress`'s progress:
```
  enum progress: {
    to_do: "to_do",
    in_progress: "in_progress",
    complete: "complete",
  }
```

db/migrate/20210614075746_add_induction_programme_choice_to_early_career_teacher_profile.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddInductionProgrammeChoiceToEarlyCareerTeacherProfile < ActiveRecord::Migration[6.1]
+  def change
+    add_column :early_career_teacher_profiles, :induction_programme_choice, :string
+  end
+end
```
kicferk1, line 7: I think the default should be `not_yet_known`. Same with mentor one.

spec/models/tracked_email_spec.rb
```diff
@@ -55,11 +56,18 @@
         expect(invite_email_ect.notify_id).to eq("test_id")
         expect(invite_email_ect.sent_to).to eq(invite_email_ect.user.email)
       end
+
+      it "does not send the email to a full induction programme ect user" do
+        ect.early_career_teacher_profile.update!(induction_programme_choice: "full_induction_programme")
```
kicferk1, line 61: Making `induction_programme_choice` an enum will let you do
```
ect.early_career_teacher_profile.full_induction_programme!
```
magic :O

app/models/early_career_teacher_profile.rb
```diff
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 
 class EarlyCareerTeacherProfile < ApplicationRecord
+  enum induction_programme_choice: {
+    core_induction_programme: "core_induction_programme",
+    full_induction_programme: "full_induction_programme",
+    not_yet_known: "not_yet_known",
+  }
```
kicferk1, line 8: Are these all options we can get? :)

Probably not. Double check with what they are in R&P - we are, after all, returning the same things they have on their model from the api.

xswoodsx: The two others are 
`design_our_own: "design_our_own", 
no_early_career_teachers: "no_early_career_teachers",`

I thought we might not want to do anything with these... If you've no teachers, then you won't need mentors and if you're designing your own then you won't have access at anytime or need it. 

Happy to add them in if you think we should. 

I had it in my head yesterday that we wouldn't get these anyway but I can't remember why... maybe it was the programme types we initially had set up didn't include these.

kicferk1: Well, whether we need them or not is a product decision. In current business logic we don't really use them.

But, from a technical perspective, it is possible that R&P will return any of those values on the user field, and crash our system when we try to save them. Because that's exactly the power and usefulness of an enum.

We should either add them to our end, or make sure that when we receive the data from R&P we cast them to things that our system understands.

How you approach it is up to you, all I want is for our app not to throw an error when it receives a valid R&P response :) You can use automated tests to simulate how the app will react to whatever you put on that field - just edit the fixture locally and see what happens when you run the tests for syncing users.

app/models/user.rb
```diff
@@ -36,9 +36,11 @@ def core_induction_programme
...
 
-  def is_on_full_induction_programme?
-    early_career_teacher_profile&.full_induction_programme? ||
-      mentor_profile&.full_induction_programme?
+  def is_not_on_core_induction_programme?
+    is_on_full_induction_programme? ||
+      is_on_designed_programme? ||
+      programme_has_no_early_career_teachers? ||
+      programme_not_yet_known?
   end
```
kicferk1: `! is_on_core_induction_programme?`?

xswoodsx: I tried this but didn't have luck, i'll have another look at it

xswoodsx: So it was causing non ect or mentor profiles to also be denied 

Doing  `!is_on_core_induction_programme? && (mentor? || early_career_teacher?)` looks to work fine

kicferk1 (approved):

PR #152, opened 2021-06-16
Add a nightly job to sync all users

kicferk1:
## Ticket and context

Currently we rely on R&P being nice, and to know what we mean by 'the user was updated' - cause we sent them `updated_since` parameter. 

If it so happens that there is some tiny change here or there (say in that user's schools's cohort's induction programme changes), that doesn't trigger a change of that user's `updated_at` field on R&P, but is important for E&L, we want to have some way of updating it.

And this is this way - nightly job to sync all users.

### Review Checks

### DIFF
--- a/app/jobs/sync_all_users_job.rb
+++ b/app/jobs/sync_all_users_job.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class SyncAllUsersJob < CronJob
+  self.cron_expression = "0 0 * * *"
+
+  queue_as :sync_users
+
+  def perform
+    Rails.logger.info "Syncing all users with Register and Partner..."
+    RegisterAndPartnerApi::SyncUsers.perform(all: true)
+  end
+end
--- a/app/services/register_and_partner_api/sync_users.rb
+++ b/app/services/register_and_partner_api/sync_users.rb
@@ -15,11 +15,15 @@ module RegisterAndPartnerApi
       none: "none",
     }.freeze
 
-    def self.perform
+    def self.perform(all: false)
       new_sync_time = Time.zone.now
       # TODO: Add pagination
       begin
-        sync_users(RegisterAndPartnerApi::User.where(filter: { updated_since: SyncUsersTimer.last_sync }))
+        if all
+          sync_users(RegisterAndPartnerApi::User.all)
+        else
+          sync_users(RegisterAndPartnerApi::User.where(filter: { updated_since: SyncUsersTimer.last_sync }))
+        end
       rescue JsonApiClient::Errors::ClientError
         # This is how the API responds when we run out of pages :/
       end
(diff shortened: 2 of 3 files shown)


app/jobs/sync_all_users_job.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class SyncAllUsersJob < CronJob
+  self.cron_expression = "0 0 * * *"
```
xswoodsx, line 4: Is this saying run the job at midnight each day?

app/services/register_and_partner_api/sync_users.rb
```diff
@@ -15,11 +15,15 @@ class SyncUsers
...
+    def self.perform(all: false)
       new_sync_time = Time.zone.now
       # TODO: Add pagination
       begin
-        sync_users(RegisterAndPartnerApi::User.where(filter: { updated_since: SyncUsersTimer.last_sync }))
+        if all
+          sync_users(RegisterAndPartnerApi::User.all)
+        else
+          sync_users(RegisterAndPartnerApi::User.where(filter: { updated_since: SyncUsersTimer.last_sync }))
+        end
```
xswoodsx, line 26: A ternary? Might be a bit long if it's on one line. Not to fussed either way, i just tend to like ternary's on a single condition

app/jobs/sync_all_users_job.rb
```diff
@@ -0,0 +1,12 @@
...
+
+  queue_as :sync_users
```
xswoodsx, line 6: What's this doing?

app/jobs/sync_all_users_job.rb
kicferk1: Yep, that is exactly what it is saying :)

app/services/register_and_partner_api/sync_users.rb
kicferk1: I prefer to use ternaries when I deal with expressions. So say
```
title = use_full? ? full_title : short_title
```
But dealing with anything that does side effects (like our `sync_users` method) is a bit scary - cause ternary expression tends to make it harder to focus on which side effects are happening when.

That said, I have an idea.

app/jobs/sync_all_users_job.rb
kicferk1: https://edgeguides.rubyonrails.org/active_job_basics.html#queues

In short, there can be many queues for our jobs. Say, one for very high priority stuff, one for sending emails, one for syncing users. And you can make those queues have different priorities - so that our worker, which is executing the queues, goes for the important things first.

We don't have many queues and prioritising in place yet, but we probably will.

xswoodsx (approved):
LGTM :)

PR #158, opened 2021-06-18
Add pagination for users api

kicferk1:
## Ticket and context

We were only fetching the first page of R&Ps users api. Not enough.

### Review Checks


xswoodsx (approved):
LGTM :) 

Out of curiosity, is there any way we could get the number of pages in the first request instead of going until we get no users returned?

kicferk1:
> LGTM :)
> 
> Out of curiosity, is there any way we could get the number of pages in the first request instead of going until we get no users returned?

Uhh, maybe. If we made the R&P api return that info. Currently I don't think it does.

PR #159, opened 2021-06-18
Add live notify key to production

kicferk1:
## Ticket and context

Ticket: We want to send some live invite emails soon.

Nope, everything is straightforward. Thrilling read of an encrypted file. If you really want to know the key, message me and I will send you the production rails credentials key.


callumacrae (approved):
why not

PR #163, opened 2021-06-21
Fix pagination 2

kicferk1:
## Ticket and context

Turns out adding pagination can be tricky. I will play around with the review app to make sure it behaves the way I expect it.

I think the real fix is removing front slashes from R&P urls. But just in case I added some more safety measures, like `sleep(1)` and checking whether we got any new R&P ids. 

### Review Checks

### DIFF
--- a/app/services/register_and_partner_api/sync_users.rb
+++ b/app/services/register_and_partner_api/sync_users.rb
@@ -20,27 +20,39 @@ module RegisterAndPartnerApi
       last_sync = SyncUsersTimer.last_sync
       base_query = all ? {} : { filter: { updated_since: last_sync } }
 
-      # is_last_page = false
-      page_number = 0
-      # until is_last_page || page_number > 10_000
-      page_number += 1
-      paginated_query = base_query.merge(page: { page: page_number, per_page: 100 })
-
-      response = RegisterAndPartnerApi::User.where(paginated_query)
-      sync_users(response)
-
-      # is_last_page = true if response.count.zero?
-      # end
+      perform_pagination(base_query)
 
       SyncUsersTimer.set_last_sync(new_sync_time)
     rescue StandardError => e
       Rails.logger.warn("Failed to sync users: #{e}")
     end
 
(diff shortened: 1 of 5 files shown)


app/services/register_and_partner_api/sync_users.rb
```diff
@@ -20,27 +20,39 @@ def self.perform(all: false)
...
       Rails.logger.warn("Failed to sync users: #{e}")
     end
 
+    def self.perform_pagination(base_query)
+      is_last_page = false
+      page_number = 0
+      all_user_ids = []
+      until is_last_page
+        page_number += 1
+        paginated_query = base_query.merge(page: { page: page_number, per_page: 10 })
```
ethanmills, line 36: Why 10 per page?

ethanmills (approved):
I'm wondering if there's an edge case here where we miss some users, if they're created at the wrong moment, or if there's some clock drift between the two servers. Do you think there would be any harm in `updated_since: last_sync - 5.minutes` just for a bit of margin?

ethanmills:
Just bear in mind that at 10 per page, 1 page per second, initial sync is going to take longer than 5 mins. Will this cause issues?

kicferk1:
> Just bear in mind that at 10 per page, 1 page per second, initial sync is going to take longer than 5 mins. Will this cause issues?

Oh, my bad, that should be 100. Good spot!

app/services/register_and_partner_api/sync_users.rb
kicferk1: Cause it made testing easier and I forgot to take it out.

kicferk1:
> I'm wondering if there's an edge case here where we miss some users, if they're created at the wrong moment, or if there's some clock drift between the two servers. Do you think there would be any harm in `updated_since: last_sync - 5.minutes` just for a bit of margin?

Eh, that's what the `all` flag is for - we sync all users at night.
I think it's pretty unlikely we miss anything this way - but concurrency is a pain.

PR #164, opened 2021-06-21
CPDEL-563: Fix importing core induction programes

kicferk1:
## Ticket and context

We were not correctly importing core induction programmes :(

### Review Checks

2. Try to log in as `rp-mentor-ambition@example.com`. You should be able to access your materials. Before this change there was no button to start.

### DIFF
--- a/app/services/register_and_partner_api/sync_users.rb
+++ b/app/services/register_and_partner_api/sync_users.rb
@@ -8,10 +8,10 @@ module RegisterAndPartnerApi
     }.freeze
 
     CIP_TYPES = {
-      ambition: "Ambition Institute",
-      edt: "Education Development Trust",
-      teach_first: "Teach First",
-      ucl: "UCL",
+      ambition: "ambition",
+      edt: "edt",
+      teach_first: "teach_first",
+      ucl: "ucl",
       none: "none",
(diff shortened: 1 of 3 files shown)


app/services/register_and_partner_api/sync_users.rb
```diff
@@ -8,10 +8,10 @@ class SyncUsers
...
 
     CIP_TYPES = {
-      ambition: "Ambition Institute",
-      edt: "Education Development Trust",
-      teach_first: "Teach First",
-      ucl: "UCL",
+      ambition: "ambition",
+      edt: "edt",
+      teach_first: "teach_first",
+      ucl: "ucl",
```
callumacrae, line 14: what is the point in this object?

callumacrae (approved):

app/services/register_and_partner_api/sync_users.rb
kicferk1: Its main goal is to avoid sending names of core induction programme between R&P and E&L - cause those names are not the same 😢 

So instead of sending names we send an enum. And this is one way you can have an enum in Ruby code. There is a carbon copy of this object in R&P.

PR #168, opened 2021-06-23
CPDEL-584 change ambition module title

xswoodsx:
## Ticket and context

Ticket:
[584](https://dfedigital.atlassian.net/jira/software/projects/CPDEL/boards/89?assignee=5f3543bb3e9e2e004db3189d&selectedIssue=CPDEL-584)

Tried this a few different ways but this seemed this most straight forward. 

### Review Checks

2. Login as admin `admin@example.com`
3. Go to the ambition modules pages and click edit module
4. Add a page header and save changes
Screenshot 2021-06-23 at 16 03 21

### DIFF
--- a/app/controllers/core_induction_programmes/modules_controller.rb
+++ b/app/controllers/core_induction_programmes/modules_controller.rb
@@ -67,7 +67,7 @@ private
   def course_module_params
     params
         .require(:course_module)
-        .permit(:ect_summary, :mentor_summary, :title, :term, :course_year_id, :previous_module_id)
+        .permit(:ect_summary, :mentor_summary, :title, :page_header, :term, :course_year_id, :previous_module_id)
   end
 
   def find_next_module
--- a/app/views/core_induction_programmes/modules/show.html.erb
+++ b/app/views/core_induction_programmes/modules/show.html.erb
@@ -4,7 +4,12 @@
     <% if policy(@course_module).edit? %>
       <%= govuk_link_to "Edit module content", edit_module_path(@course_module), button: true%>
     <% end %>
-    <h1 class="govuk-heading-l"> <%= @course_module.term_and_title %></h1>
+
+    <% if @course_module.page_header? %>
+      <h1 class="govuk-heading-l"><%= @course_module.page_header %></h1>
+    <% else %>
+      <h1 class="govuk-heading-l"><%= @course_module.term_and_title %></h1>
+    <% end %>
 
     <ul class="app-task-list__items">
       <% if @course_lessons_with_progress.each do |lesson| %>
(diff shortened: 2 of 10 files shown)


app/controllers/core_induction_programmes/modules_controller.rb
```diff
@@ -79,4 +80,8 @@ def find_next_module
   def fill_data_layer
     data_layer.add_module_info(@course_module)
   end
+
+  def cip_is_ambition?
+    @course_module.course_year.core_induction_programme.name == "Ambition Institute"
+  end
```
kicferk1, line 86: I do not want to open that can of worms.

Instead of making our code aware of 'oh, you are displaying ambition, display it differently', please add a field called `page_header` to all modules, make it nullable, and in place of the page title display that - unless it is empty, then display the title.

app/views/core_induction_programmes/modules/show.html.erb
```diff
@@ -4,7 +4,12 @@
     <% if policy(@course_module).edit? %>
       <%= govuk_link_to "Edit module content", edit_module_path(@course_module), button: true%>
     <% end %>
-    <h1 class="govuk-heading-l"> <%= @course_module.term_and_title %></h1>
+
+    <% if @cip_is_ambition %>
+      <h1 class="govuk-heading-l"> <%= "#{@course_module.term.capitalize} term: #{@course_module.title}" %></h1>
+    <% else %>
+      <h1 class="govuk-heading-l"> <%= @course_module.term_and_title %></h1>
+    <% end %>
```
kicferk1, line 12: This has a potential of turning pretty horrific, when we eventually add a few more rules for a few more providers. Avoiding this complexity is why we should move this logic into its own human editable database field.

app/views/core_induction_programmes/modules/show.html.erb
```diff
@@ -5,10 +5,10 @@
       <%= govuk_link_to "Edit module content", edit_module_path(@course_module), button: true%>
     <% end %>
 
-    <% if @cip_is_ambition %>
-      <h1 class="govuk-heading-l"> <%= "#{@course_module.term.capitalize} term: #{@course_module.title}" %></h1>
+    <% if @course_module.page_header? %>
```
kicferk1, line 8: What happens when `page_header` is an empty string?
I expected `@course_module.page_header.present?`, but maybe that is enough :)

app/views/core_induction_programmes/modules/show.html.erb
xswoodsx: V.good point!

Hopefully the changes should mean more provider specific changes aren't as terrible to do.

kicferk1 (approved):

app/views/core_induction_programmes/modules/show.html.erb
xswoodsx: Yeah looks to be enough. Adding an empty string doesn't change the title , it goes to term_and_title

PR #172, opened 2021-06-24
URL changes

callumacrae:
🐉🤠🐉🤠🐉🤠🐉🤠🐉🤠🐉🤠

(work in progress)


app/controllers/core_induction_programmes/mentor_materials_controller.rb
```diff
@@ -1,11 +1,13 @@
 # frozen_string_literal: true
 
-class CoreInductionProgrammes::MentorMaterialsController < ApplicationController
+class CoreInductionProgrammes::MentorMaterialsController < CoreInductionProgrammes::LessonsController
```
kicferk1, line 3: Are we doing all of this controller inheritance to have access to `load_xxx` methods that are shared between them?

If so, I would really prefer those load methods to be in a helper, and include that helper in all relevant controllers. That saves us the need to worry about what actions we need to skip, and what needless methods we inherit.

db/migrate/20210624134755_add_slug_to_cips.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+SLUGS = {
+  "Ambition Institute": "ambition",
+  "Education Development Trust": "edt",
+}.freeze
```
kicferk1, line 6: Probably need more

db/schema.rb
```diff
@@ -42,6 +42,7 @@
     t.datetime "updated_at", precision: 6, null: false
     t.uuid "course_year_one_id"
     t.uuid "course_year_two_id"
+    t.string "slug"
```
kicferk1, line 45: I would expect this to cause some changes in the seed dump, but not 400 lines. Before we merge it, could you make sure you get latest data from develop (on old db version), migrate it to add the slug and run `rails cip_seed_dump` afterwards? I hope for like 10 line diff (just for core induction programmes) :)

db/migrate/20210624134755_add_slug_to_cips.rb
callumacrae: UCL -> ucl
Teach First -> teach-first

are both basically what we want already so no need to override them

db/schema.rb
callumacrae: yea i'm pretty sure most of the diff is the changes you made last week but backwards - i'll update from develop at some point :)

callumacrae:
note to self: i have somehow absolutely destroyed the seed dump file

PR #173, opened 2021-06-25
CPDEL-593: Add an option to force-send emails

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDEL-593

### Review Checks

1. You can't.

### DIFF
--- a/app/models/invite_email_mentor.rb
+++ b/app/models/invite_email_mentor.rb
@@ -7,8 +7,9 @@ class InviteEmailMentor < TrackedEmail
     UserMailer.mentor_welcome_email(user)
   end
 
-  def send!
-    if user.is_on_core_induction_programme? && Time.zone.now >= INVITES_SENT_FROM
+  def send!(force_send: false)
+    can_send_automatically = user.is_on_core_induction_programme? && Time.zone.now >= INVITES_SENT_FROM
+    if can_send_automatically || force_send
       super
     end
   end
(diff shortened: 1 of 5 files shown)


app/models/invite_email_mentor.rb
```diff
@@ -8,7 +8,9 @@ def mail_to_send
   end
 
   def send!(force_send: false)
-    can_send_automatically = user.is_on_core_induction_programme? && Time.zone.now >= INVITES_SENT_FROM
+    return unless user.is_on_core_induction_programme?
+
+    can_send_automatically = Time.zone.now >= INVITES_SENT_FROM
```
xswoodsx: Was the force send sending non-cip users an email before making this change?

kicferk1: It in principle could, yes.

kicferk1: And I got worried that I would force-send emails to a wrong user by accident :P

app/models/tracked_email.rb
```diff
@@ -15,7 +15,7 @@ def mail_to_send
     raise NotImplementedError
   end
 
-  def send!
+  def send!(*)
```
xswoodsx, line 18: This is splat operator/naked splat right? It lets you pass any number of arguments to super but we aren't calling super at any point are we? 

I'm just trying to understand why it's needed if we're overriding the method in invite_email_ect and invite_email_mentor .

xswoodsx (approved):
LGTM :)

PR #180, opened 2021-06-30
Automatically handle generating cip_seed.rb as well as cip_seed_dump.rb

callumacrae:
## Ticket and context

vim wasn't a fan of doing this in the editor so i wrote a script for it - figured i'd add it to the rake task

why haven't i had to double escape the square brackets? i was expecting ruby to get rid of the escape characters

## Product review

this is entire developer experience stuff and doesn't affect the actual product :)


kicferk1 (approved):
I ran it locally and it worked, looks good to me :)

kicferk1 (commented):
Although maybe let's not merge it just yet - I wanna see how review app behaves with it :)

kicferk1:
And it seems it works 🚢

PR #185, opened 2021-07-01
Add a rake task to invite mentors with unsent invites

kicferk1:
## Ticket and context

I want to send invite emails to all mentors who are on our system. The syncing script made an email for each of them. And this will send them. If they are doing FIP, the `send!` method shouldn't work. 

### DIFF
--- a/lib/tasks/invite_mentors.rake
+++ b/lib/tasks/invite_mentors.rake
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+namespace :invites do
+  desc "Invite mentor users"
+  task send_mentor_invites: :environment do
+    puts("Sending mentor invites...")
+
+    unsent_emails = InviteEmailMentor.where(sent_at: nil)
+    unsent_emails.each do
+      email.send!(force_send: true)
+    rescue Notifications::Client::RateLimitError
+      sleep(1)
+      email.send!(force_send: true)
+    rescue StandardError
+      puts("Error sending email, id: #{email&.id} ... skipping")
+    end
+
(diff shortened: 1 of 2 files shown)


lib/tasks/invite_mentors.rake
```diff
@@ -0,0 +1,20 @@
...
+namespace :invites do
+  desc "Invite mentor users"
+  task send_mentor_invites: :environment do
+    puts("Sending mentor invites...")
+
+    unsent_emails = InviteEmailMentor.where(sent_at: nil)
+    unsent_emails.each do
+      email.send!(force_send: true)
+    rescue Notifications::Client::RateLimitError
+      sleep(1)
```
xswoodsx, line 12: sleeps the equivalent of setInterval in JS? Why's it needed here? Is it incase we're sending them to quickly after each other and this just adds a delay to allow the task to continue without failing completely? If that's right, what's the rate limit?

xswoodsx: 360 per second by the looks of it

xswoodsx (approved):

kicferk1:
Apparently we don't need this just yet - we want to send the emails to a limited subset of our mentors due to DQT validation concerns (it is coming down the line and we don't want to have to do it for too many people).

lib/tasks/invite_mentors.rake
kicferk1: Yeah, that is what it does. I think 1 second is enough, I stole the idea from R&P who have a lot more experience sending vast amounts of emails.

PR #191, opened 2021-07-05
Fix sign in bug

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDEL-631

The bug was caused by some lovely inner machinations of devise. I think `current_user` in some edge case was signing the user in - say, if there is no cached saved user and warden't `authenticate!` returned it a user.

So the solution is to not use warden's strategies for this stuff - I got pretty fed up trying to fix it without giving up on the strategy but it just was not very nice :/ 

But if you don't use any strategies, devise stops being nice - it doesn't draw the right routes, it doesn't redirect to sign-in when user is not authenticated on some pages, and so on... so I added a dummy strategy to keep it happy :( 

### Review Checks

2. Try signing in as `cip-less-ect@example.com`.
3. After you see error message, click on the service name.
4. You should not see the dashboard.


app/controllers/users/sessions_controller.rb
```diff
@@ -51,6 +55,34 @@ def login_token_expired?
...
+    )
+
+    if result
+      url = Rails.application.routes.url_helpers.users_confirm_sign_in_url(
+        login_token: user.login_token,
+        host: Rails.application.config.domain,
+        **UtmService.email(:sign_in),
+      )
+
+      UserMailer.sign_in_email(user: user, url: url, token_expiry: token_expiry.localtime.to_s(:time)).deliver_now
```
xswoodsx, line 82: Why did the mailer need requiring in for the old strategy but not here?

lib/devise/strategies/passwordless_authenticatable.rb
```diff
@@ -1,63 +0,0 @@
...
-module Devise
-  module Strategies
-    class PasswordlessAuthenticatable < Authenticatable
-      TEST_USERS = %w[admin@example.com].freeze
-
-      class Error < StandardError; end
-
-      class EmailNotFoundError < Error; end
-
-      class LoginIncompleteError < Error; end
```
xswoodsx, line 15: How come these errors haven't needed to be moved over as well?

app/controllers/users/sessions_controller.rb
```diff
@@ -3,18 +3,22 @@
...
   before_action :redirect_to_dashboard, only: %i[sign_in_with_token redirect_from_magic_link]
   before_action :ensure_login_token_valid, only: %i[sign_in_with_token redirect_from_magic_link]
+  TEST_USERS = %w[admin@example.com].freeze
 
   def create
     self.resource = validate_email
+
     if resource.errors.any?
-      render :new
+      render :new and return
```
xswoodsx, line 12: What does the 'and return' add to this? i think render was returning that view or does return mean something else in this instance?

kicferk1: I added an extra 'branch' to our logic - `user_requires_magic_link`.

And it became apparent that the whole method will read nicer if we use an early return.

app/controllers/users/sessions_controller.rb
kicferk1: Rails magic :D

xswoodsx (approved):

PR #196, opened 2021-07-06
Remove unused fields from db

kicferk1:
## Ticket and context

No ticket, I just don't like them floating around. 

Includes a sneaky fix for splitting mentor material parts. 

### Review Checks

### DIFF
--- a/app/models/mentor_material.rb
+++ b/app/models/mentor_material.rb
@@ -3,13 +3,12 @@
 class MentorMaterial < ApplicationRecord
   include OrderHelper
 
-  # TODO: hopefully remove some of those
-  belongs_to :core_induction_programme, optional: true
-  belongs_to :course_year, optional: true
-  belongs_to :course_module, optional: true
   belongs_to :course_lesson
+  has_one :course_module, through: :course_lesson
+  has_one :course_year, through: :course_module
+  delegate :core_induction_programme, to: :course_year
 
-  has_many :mentor_material_parts
+  has_many :mentor_material_parts, dependent: :destroy
 
   validates :title, presence: { message: "Enter a title" }, length: { maximum: 255 }
 
--- a/app/models/mentor_material.rb
+++ b/app/models/mentor_material.rb
@@ -12,18 +12,6 @@ class MentorMaterial < ApplicationRecord
 
   validates :title, presence: { message: "Enter a title" }, length: { maximum: 255 }
 
-  def get_core_induction_programme
-    if course_lesson.present?
-      course_lesson.course_year.core_induction_programme
-    elsif course_module.present?
-      course_module.course_year.core_induction_programme
-    elsif course_year.present?
-      course_year.core_induction_programme
-    else
-      core_induction_programme
-    end
-  end
-
   def mentor_material_parts_in_order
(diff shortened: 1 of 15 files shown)


app/models/mentor_material.rb
```diff
@@ -3,13 +3,12 @@
...
   include OrderHelper
 
-  # TODO: hopefully remove some of those
-  belongs_to :core_induction_programme, optional: true
-  belongs_to :course_year, optional: true
-  belongs_to :course_module, optional: true
   belongs_to :course_lesson
+  has_one :course_module, through: :course_lesson
+  has_one :course_year, through: :course_module
+  delegate :core_induction_programme, to: :course_year
```
xswoodsx, line 9: whilst you've changed this to delegate cip to course_year can we change the get_core_induction_programme method and just return course_year.core_induction programme?

xswoodsx, line 9: As you've done this can we need the get_core_induction_programme method? or can be it reduced just to return the cip or its id?

app/views/core_induction_programmes/mentor_materials/index.html.erb
```diff
@@ -11,10 +11,7 @@
...
       <% cip.mentor_materials.each do |mentor_material| %>
         <p><%= govuk_link_to mentor_material.title, mentor_material_path(mentor_material) %></p>
-        <% if mentor_material.course_lesson %>
-          <p>Corresponding lesson: <%= govuk_link_to mentor_material.course_lesson.title, lesson_path(mentor_material.course_lesson) %></p>
-          <% else %> <p>No corresponding lesson found</p>
-       <%end %>
+        <p>Corresponding lesson: <%= govuk_link_to mentor_material.course_lesson.title, lesson_path(mentor_material.course_lesson) %></p>
       <% end %>
       <% else %>
       <p>No mentor materials were found!</p>
```
xswoodsx, line 17: Would it be worth just removing index completely? I know i've not touched it since we added links elsewhere

kicferk1: I touched it quite a few times - whenever there is a ticket saying 'do something for each mentor material for a particular CIP'.

spec/models/mentor_material_spec.rb
```diff
@@ -5,9 +5,8 @@
 RSpec.describe MentorMaterial, type: :model do
   describe "associations" do
     it { is_expected.to belong_to(:course_lesson) }
-    it { is_expected.to belong_to(:course_module).optional }
-    it { is_expected.to belong_to(:course_year).optional }
-    it { is_expected.to belong_to(:core_induction_programme).optional }
+    it { is_expected.to have_one(:course_module) }
+    it { is_expected.to have_one(:course_year) }
```
xswoodsx, line 9: Should we test for the delegate or is that not typically done?

app/models/mentor_material.rb
kicferk1: Good spot!

spec/models/mentor_material_spec.rb
kicferk1: Probably worth it

xswoodsx (approved):
Looks good :)

PR #197, opened 2021-07-06
Update the way we store course year to cip relations

kicferk1:
## Ticket and context

I think it is pretty annoying to try to find the right cip on a year from two relations. I think it will be easier to handle things if we go back towards course year belonging to a core induction programme, with an integer position. 

This is part one, which has to be released across all environments first. Part two will remove the unused fields and code. We can't remove them just yet, cause migrating things would be a lot harder that way. 

### Review Checks

### DIFF
--- a/app/models/course_year.rb
+++ b/app/models/course_year.rb
@@ -7,6 +7,7 @@ class CourseYear < ApplicationRecord
   has_one :core_induction_programme_one, class_name: "CoreInductionProgramme", foreign_key: :course_year_one_id
   has_one :core_induction_programme_two, class_name: "CoreInductionProgramme", foreign_key: :course_year_two_id
   belongs_to :core_induction_programme
+  acts_as_list scope: :core_induction_programme
   has_many :course_modules, dependent: :delete_all
   has_many :mentor_materials
 
@@ -49,7 +50,7 @@ class CourseYear < ApplicationRecord
   end
 
   def to_param
-    "year-#{self == core_induction_programme.course_year_one ? '1' : '2'}"
+    "year-#{position}"
   end
 
 private
--- a/app/models/course_year.rb
+++ b/app/models/course_year.rb
@@ -6,6 +6,7 @@ class CourseYear < ApplicationRecord
 
   has_one :core_induction_programme_one, class_name: "CoreInductionProgramme", foreign_key: :course_year_one_id
   has_one :core_induction_programme_two, class_name: "CoreInductionProgramme", foreign_key: :course_year_two_id
+  belongs_to :core_induction_programme
   has_many :course_modules, dependent: :delete_all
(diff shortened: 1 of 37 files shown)


db/seeds/cip_seed.rb
```diff
@@ -1,18 +1,18 @@
...
-  ["a4dc302c-ab71-4d7b-a10a-3116a778e8d5", "Your module materials", "<p></p>", "Coaching and mentoring support materials"],
-  ["793ad37c-09c1-49a4-922f-78b29be4bc87", "Your module materials", "TBC", "Coaching and mentoring support materials"],
-  ["fbf47b24-691b-4114-af34-6adcc9b45346", "Your module materials", "<p></p>", "Coaching and mentoring support materials"],
-  ["b61a7399-c503-4e37-81af-e45bdaddd93c", "Your module materials", "<p></p>", "Coaching and mentoring support materials"],
-  ["41a0ed38-4a04-436c-8559-2b7fc3b917c6", "Your module materials", "<p></p>", "Coaching and mentoring support materials"],
-  ["d7ac2099-81cb-4f95-9c32-ca68775c61d0", "Your module materials", "<p></p>", "Coaching and mentoring support materials"],
-  ["085b64e0-bd12-4795-a5b9-c3944af9d02e", "Your module materials", "TBC", "Coaching and mentoring support materials"],
-  ["441979d3-787c-42bd-9072-2309125ef709", "Your module materials", "<p></p>", "Coaching and mentoring support materials"]
-], on_duplicate_key_update: { conflict_target: [:id], columns: :all })
 CoreInductionProgramme.import([:id, :name, :course_year_one_id, :course_year_two_id, :slug], [
```
xswoodsx, line 1: Do the course_year_one_id and course_year_two_ids need to be in there, along with the nils below?

kicferk1: I will remove them in a later PR. I think they don't need to be here, but I prefer to cut them when I cut them out of the database.

app/models/course_year.rb
```diff
@@ -41,18 +43,14 @@ def modules_with_progress(user, modules_to_order = course_modules)
...
-
   def title_for(user)
     return title if mentor_title.blank?
 
     user.mentor? ? mentor_title : title
   end
 
   def to_param
-    "year-#{self == core_induction_programme.course_year_one ? '1' : '2'}"
+    "year-#{position}"
```
xswoodsx, line 53: where is position coming from? is it the acts as list that returns it?

kicferk1: It is a field in a database, a migration is adding it :)

xswoodsx: ah yeah, hadn't spotted that column

xswoodsx (approved):
LGTM :)

PR #199, opened 2021-07-07
Increase number of service instances

kicferk1:
## Ticket and context

We need to bump our number of services on live environment, and change the db plan to high availability.

I intend to use this review app to test changing the db plan when we have some data, that's why I removed safe reset from the review command. I will seed it manually, make some changes, push a change that will change the db plan, and see if it works. I hope it does. 


kicferk1:
Cannot update to use different encryption settings, will do round 2 in another PR

PR #223, opened 2021-07-28
Small fixes

kicferk1:
## Ticket and context

We need some fixes which I found out about recently.


ethanmills (approved):
SHA appears in healthcheck

PR #228, opened 2021-08-04
Cpdel 708 guidance speed bump

xswoodsx:
## Ticket and context

Ticket: 

[708](https://dfedigital.atlassian.net/jira/software/projects/CPDEL/boards/89?selectedIssue=CPDEL-708)
[765](https://dfedigital.atlassian.net/jira/software/projects/CPDEL/boards/89?selectedIssue=CPDEL-765)

708 is the bulk of the change implemented. 765 is just the addition of a button on guidance pages and rewording a link. 

I found raising an error when no radio is selected a pain. Passing true or false as a value meant that a radio button was preselected, which we don't want. 

Naming, as always, is a pain so welcome any thoughts on naming appropriately. 

### Review Checks

2. login is a mentor or ect 
3. Click start now on the dashboard. 
4. Go through the multiple scenarios set out in the acceptance criteria. 
5. You should only see the page once for each user type so you'll need to log in to another example account to try again.

### DIFF
--- a/app/controllers/core_induction_programmes/core_induction_programmes_controller.rb
+++ b/app/controllers/core_induction_programmes/core_induction_programmes_controller.rb
@@ -13,6 +13,8 @@ class CoreInductionProgrammes::CoreInductionProgrammesController < ApplicationCo
   end
 
   def show
+    redirect_to guidance_question_path and return if current_user&.participant_profile&.show_guidance_speedbump?
+
     data_layer.add_cip_info(@core_induction_programme)
     authorize @core_induction_programme
     redirect_to cip_year_path(@core_induction_programme, @core_induction_programme.course_years.first)
--- a/app/controllers/core_induction_programmes/core_induction_programmes_controller.rb
+++ b/app/controllers/core_induction_programmes/core_induction_programmes_controller.rb
@@ -13,8 +13,6 @@ class CoreInductionProgrammes::CoreInductionProgrammesController < ApplicationCo
   end
 
   def show
-    redirect_to guidance_question_path and return if current_user&.participant_profile&.show_guidance_speedbump?
-
     data_layer.add_cip_info(@core_induction_programme)
     authorize @core_induction_programme
     redirect_to cip_year_path(@core_induction_programme, @core_induction_programme.course_years.first)
--- a/app/controllers/training_and_support_controller.rb
+++ b/app/controllers/training_and_support_controller.rb
@@ -17,14 +17,11 @@ class TrainingAndSupportController < ApplicationController
   def update
     @guidance_speedbump_form = GuidanceSpeedbumpForm.new(guidance_question_params)
 
-    if @guidance_speedbump_form.view_guidance_option.blank?
-      @guidance_speedbump_form.errors.add :view_guidance_option, "Select if you would like a brief summary of the programme"
-      render :guidance_question and return
-    end
+    render :guidance_question and return unless @guidance_speedbump_form.valid?
 
     current_user.participant_profile.update!(guidance_seen: true)
 
-    if view_guidance?(@guidance_speedbump_form)
+    if @guidance_speedbump_form.view_guidance?
       redirect_to training_and_support_path
     else
       redirect_to cip_path(current_user.core_induction_programme)
@@ -36,10 +33,6 @@ class TrainingAndSupportController < ApplicationController
   end
 
 private
-
-  def view_guidance?(guidance_speedbump_form)
-    guidance_speedbump_form.view_guidance_option == "view_guidance"
-  end
 
   def guidance_question_params
     params.fetch(:guidance_speedbump_form, {}).permit(:view_guidance_option)
--- a/app/controllers/training_and_support_controller.rb
+++ b/app/controllers/training_and_support_controller.rb
@@ -10,20 +10,21 @@ class TrainingAndSupportController < ApplicationController
       raise Pundit::NotAuthorizedError, "This page is only available to CIP participants"
     end
 
+    current_user.participant_profile.update!(guidance_seen: true)
     @provider_name_key = @provider.name.downcase.gsub(/ /, "_")
   end
 
   def update
-    @participant = current_user.participant_profile
+    @guidance_speedbump_form = GuidanceSpeedbumpForm.new(guidance_question_params)
 
-    if params[@participant.model_name.param_key].blank?
-      @participant.errors.add :viewed_guidance, "Select if you would like a brief summary of the programme"
+    if @guidance_speedbump_form.view_guidance_option.blank?
+      @guidance_speedbump_form.errors.add :view_guidance_option, "Select if you would like a brief summary of the programme"
       render :guidance_question and return
     end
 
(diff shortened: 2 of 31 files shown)


db/migrate/20210803194547_add_show_guidance_speedbump_to_early_career_teacher_profile.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddShowGuidanceSpeedbumpToEarlyCareerTeacherProfile < ActiveRecord::Migration[6.1]
+  def change
+    add_column :early_career_teacher_profiles, :show_guidance_speedbump, :boolean, default: true
```
kicferk1, line 5: Should we name the flag something less intent-ful and more state-ful?

`guidance_seen`? perhaps? 

That leads to another interesting point. If the user sees the guidance some other way than speedbump, should we still show them a speedbump?

app/controllers/training_and_support_controller.rb
```diff
@@ -12,4 +12,32 @@ def show
...
+
+  def update
+    @participant = current_user.participant_profile
+
+    if params[@participant.model_name.param_key].blank?
+      @participant.errors.add :viewed_guidance, "Select if you would like a brief summary of the programme"
+      render :guidance_question and return
+    end
+
+    @participant.update!(show_guidance_speedbump: false)
```
kicferk1, line 24: I wonder if we should do the same in `show` action. After all, the intent is that the uses rees the guidance before they go to materials. If they happen to see them via a link on the dashboard, that probably counts, right?

app/controllers/core_induction_programmes/core_induction_programmes_controller.rb
```diff
@@ -13,6 +13,8 @@ def index
   end
 
   def show
+    redirect_to guidance_question_path and return if current_user&.participant_profile&.show_guidance_speedbump?
+
```
kicferk1, line 17: I would be tempted to keep this logic unchanged, and change where the link goes on the dashboard, depending on the flag in the db.

Being redirected immediately when you go to a page makes for slightly worse UX, I think.

app/controllers/training_and_support_controller.rb
xswoodsx: In short, we still want to show them the speedbump.

A notes been put on the ticket to say - a user may already click on the guidance separately, ahead of seeing the speedbump. However, we will show them the speedbump guidance regardless of whether they have independently clicked on the guidance straight from the dashboard or not.

db/migrate/20210803194547_add_show_guidance_speedbump_to_early_career_teacher_profile.rb
xswoodsx: I went with something similar originally and then swapped to the above, as `guidance_seen` could be true before they've hit the speedbump. 

I've gone through a bunch of different names, all not that great

How about... `given_view_guidance_option? `

app/controllers/training_and_support_controller.rb
kicferk1: That... feels awkward. I would double check that this is a product intent, rather than a perceived technology limitation.

db/migrate/20210803194547_add_show_guidance_speedbump_to_early_career_teacher_profile.rb
kicferk1: `guidance_speedbump_seen`?

app/controllers/training_and_support_controller.rb
xswoodsx: No problem, i'll raise it at stand up

db/migrate/20210803194547_add_show_guidance_speedbump_to_early_career_teacher_profile.rb
xswoodsx: Depending whether we decide to still show the speedbump or not i'll go with either `guidance_seen` or `guidance_speedbump_seen`

xswoodsx: We're going with `guidance_seen`

app/controllers/training_and_support_controller.rb
```diff
@@ -10,34 +10,38 @@ def show
...
   end
 
   def update
-    @participant = current_user.participant_profile
+    @guidance_speedbump_form = GuidanceSpeedbumpForm.new(guidance_question_params)
 
-    if params[@participant.model_name.param_key].blank?
-      @participant.errors.add :viewed_guidance, "Select if you would like a brief summary of the programme"
+    if @guidance_speedbump_form.view_guidance_option.blank?
+      @guidance_speedbump_form.errors.add :view_guidance_option, "Select if you would like a brief summary of the programme"
```
kicferk1: This can be done by adding `validates :view_guidance_option, presence: { message: "Select if you would like a brief summary of the programme" }` in the form. And then here, you would say:

```
render :guidance_question and return unless @guidance_speedbump_form.valid?

# redirect code
```

app/controllers/training_and_support_controller.rb
```diff
@@ -10,34 +10,38 @@ def show
...
   end
 
 private
 
-  def view_guidance?
-    participant = current_user.participant_profile
-    params[participant.model_name.param_key][:show_guidance_speedbump] == "view guidance"
+  def view_guidance?(guidance_speedbump_form)
+    guidance_speedbump_form.view_guidance_option == "view_guidance"
+  end
```
kicferk1, line 42: Maybe this wants to be a method on the form? so in your controller you would write `@guidance_speedbump_form.view_guidance?`, and in your form
```
  def view_guidance?
    view_guidance_option == "view_guidance"
  end
```

app/views/training_and_support/guidance_question.html.erb
```diff
@@ -2,16 +2,16 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= form_for @participant, url: guidance_question_path, method: :post do |form| %>
+    <%= form_with model: @guidance_speedbump_form, url: guidance_question_path, method: :post do |form| %>
```
kicferk1, line 5: Remember to add form id so Chris can track it :)

kicferk1 (commented):
looks better, a few small comments :)

app/controllers/training_and_support_controller.rb
xswoodsx: Thanks for this, i did try to validate it this way but was missing the valid method in the controller. I just assumed Rails would pick it up.

app/controllers/training_and_support_controller.rb
xswoodsx: Yeah, makes sense, will move it.

app/views/training_and_support/guidance_question.html.erb
xswoodsx: Good spot! Didn't think about the id in there

kicferk1 (approved):

PR #237, opened 2021-08-20
Cpdel 720 update onboarding email check

xswoodsx:
## Ticket and context

Ticket: [720](https://dfedigital.atlassian.net/jira/software/projects/CPDEL/boards/89?selectedIssue=CPDEL-720)

We've removed the date restriction on the invite email to users with a featureflag. There's some changes in the terraform workspace variables. **Prod and review-apps** have the `EMAIL_SENDING_ENABLED: false`, where as **staging and deployed development** are set too true. 

### Review Checks

### DIFF
--- a/spec/services/invite_participants_spec.rb
+++ b/spec/services/invite_participants_spec.rb
@@ -47,11 +47,11 @@ RSpec.describe InviteParticipants do
       end
 
       it "does not create an email for ect" do
-        expect { InviteParticipants.run([ect.email, mentor.email]) }.to change { InviteEmailEct.count }.by(0)
+        expect { InviteParticipants.run([ect.email, mentor.email]) }.not_to(change { InviteEmailMentor.count })
       end
 
       it "does not create an email for mentor" do
-        expect { InviteParticipants.run([ect.email, mentor.email]) }.to change { InviteEmailMentor.count }.by(0)
+        expect { InviteParticipants.run([ect.email, mentor.email]) }.not_to(change { InviteEmailMentor.count })
       end
     end
 
--- a/spec/services/invite_participants_spec.rb
+++ b/spec/services/invite_participants_spec.rb
@@ -46,12 +46,12 @@ RSpec.describe InviteParticipants do
         InviteEmailMentor.create!(user: mentor, sent_at: Time.zone.now)
       end
 
-      it "creates an email for ect" do
-        expect { InviteParticipants.run([ect.email, mentor.email]) }.to change { InviteEmailEct.count }.by(1)
+      it "does not create an email for ect" do
+        expect { InviteParticipants.run([ect.email, mentor.email]) }.not_to change { InviteEmailEct.count }
       end
 
-      it "creates an email for mentor" do
-        expect { InviteParticipants.run([ect.email, mentor.email]) }.to change { InviteEmailMentor.count }.by(1)
+      it "does not create an email for mentor" do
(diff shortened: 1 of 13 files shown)


terraform/workspace-variables/production_app_env.yml
```diff
@@ -11,3 +11,5 @@ REGISTER_AND_PARTNER_URL: "https://manage-training-for-early-career-teachers.edu
 DOMAIN: support-for-early-career-teachers.education.gov.uk
 GOVUK_WEBSITE_ROOT: support-for-early-career-teachers.education.gov.uk
 GOVUK_APP_DOMAIN: support-for-early-career-teachers.education.gov.uk
+EMAIL_SENDING_ENABLED: false
```
kicferk1, line 14: Hmmmmmm. This looks like a feature flag. Is it?

If so, I am pretty sure blue-green deploys that we are using will blow away any changes you make to enviroment after an app is deployed. So environment is not the best place for it. 

R&P had a problem like that not too long ago, and they decided to persist feature flags in their database. 

Check out https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/services/feature_flag.rb.

Alternatively, I guess you could activate it in prod by pushing a release changing this value to true. And maybe that is god enough for your needs.

xswoodsx: Yeah, it's a feature flag. I think the idea is to activate it in prod by pushing a release as you've said. 

We had a look at the feature flag set up in R&P but decided to just go with a quicker option as it's just this one flag. I also had a look at the way it's implemented in [find teacher training](https://github.com/DFE-Digital/find-teacher-training/search?q=feature_flags), unless this creates the same problem you mentioned?

kicferk1: I think if you are happy with managing it with prod releases then the current implementation should work just fine.

FriedSock: @kicferk1 Yeah this seemed like the simplest way we could have settings we could easily change between different environments (I guess we could also have used the credentials files). I think its fine to have to make a PR to change these values since we only have one "flag" for now 🤷

spec/services/invite_participants_spec.rb
```diff
@@ -46,12 +46,12 @@
...
-      it "creates an email for ect" do
-        expect { InviteParticipants.run([ect.email, mentor.email]) }.to change { InviteEmailEct.count }.by(1)
+      it "does not create an email for ect" do
+        expect { InviteParticipants.run([ect.email, mentor.email]) }.to change { InviteEmailEct.count }.by(0)
       end
 
-      it "creates an email for mentor" do
-        expect { InviteParticipants.run([ect.email, mentor.email]) }.to change { InviteEmailMentor.count }.by(1)
+      it "does not create an email for mentor" do
+        expect { InviteParticipants.run([ect.email, mentor.email]) }.to change { InviteEmailMentor.count }.by(0)
```
kicferk1, line 54: Above we use 
```
expect { InviteParticipants.run([ect.email, mentor.email]) }.not_to(change { InviteEmailMentor.count })
```
will it not work here?

kicferk1 (approved):

spec/services/invite_participants_spec.rb
xswoodsx: No... rubocop threw a wobbler and it broke the linter.

PR #241, opened 2021-08-24
[CPDEL-713] CIP change alert

FriedSock:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDEL-713

### Review Checks

### DIFF
--- a/app/controllers/cip_change_alert_controller.rb
+++ b/app/controllers/cip_change_alert_controller.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+class CipChangeAlertController < ApplicationController
+  before_action :authenticate_user!
+
+  def start; end
+
+  def acknowledge_start
+    cip_change_message&.destroy!
+    redirect_to cip_path(current_user.core_induction_programme)
+  end
+
+  def guidance; end
+
+  def acknowledge_guidance
+    cip_change_message&.destroy!
+    redirect_to training_and_support_path
+  end
+
+private
+
+  def original_cip_name
+    cip_change_message.original_cip.name
+  end
+
+  def new_cip_name
+    cip_change_message.new_cip.name
+  end
+
+  helper_method :original_cip_name, :new_cip_name
+
+  def cip_change_message
+    @cip_change_message ||= current_user.cip_change_message
+  end
+end
--- a/app/helpers/dashboard_helper.rb
+++ b/app/helpers/dashboard_helper.rb
@@ -11,7 +11,7 @@ module DashboardHelper
 
   def start_button_path(user)
     if user.cip_change_message.present?
-      cip_change_alert_start_path
+      user.participant_profile.guidance_seen? ? cip_change_alert_start_path : cip_path(user.core_induction_programme)
     else
       user.participant_profile.guidance_seen? ? cip_path(user.core_induction_programme) : guidance_question_path
     end
@@ -24,4 +24,4 @@ module DashboardHelper
       training_and_support_path
     end
   end
- end
+end
--- a/app/helpers/dashboard_helper.rb
+++ b/app/helpers/dashboard_helper.rb
@@ -8,4 +8,12 @@ module DashboardHelper
       "Start now"
     end
   end
-end
+
+  def start_button_path(user)
+    user.participant_profile.guidance_seen? ? cip_path(user.core_induction_programme) : guidance_question_path
+  end
+
+  def how_this_programme_works_path(_user)
+    training_and_support_path
+  end
+ end
--- a/app/helpers/dashboard_helper.rb
+++ b/app/helpers/dashboard_helper.rb
@@ -10,10 +10,18 @@ module DashboardHelper
   end
 
   def start_button_path(user)
-    user.participant_profile.guidance_seen? ? cip_path(user.core_induction_programme) : guidance_question_path
+    if user.cip_change_message.present?
+      cip_change_alert_start_path
+    else
+      user.participant_profile.guidance_seen? ? cip_path(user.core_induction_programme) : guidance_question_path
+    end
   end
 
-  def how_this_programme_works_path(_user)
-    training_and_support_path
+  def how_this_programme_works_path(user)
+    if user.cip_change_message.present?
+      cip_change_alert_guidance_path
+    else
+      training_and_support_path
+    end
   end
  end
--- a/app/helpers/dashboard_helper.rb
+++ b/app/helpers/dashboard_helper.rb
@@ -11,7 +11,7 @@ module DashboardHelper
 
   def start_button_path(user)
     if user.cip_change_message.present?
-      user.participant_profile.guidance_seen? ? cip_change_alert_start_path : cip_path(user.core_induction_programme)
+      cip_change_alert_start_path
     else
       user.participant_profile.guidance_seen? ? cip_path(user.core_induction_programme) : guidance_question_path
     end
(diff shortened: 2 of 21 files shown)


app/controllers/cip_change_alert_controller.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+class CipChangeAlertController < ApplicationController
+  before_action :authenticate_user!
+
+  def start
+    @original_cip_name = cip_change_message.original_cip.name
+    @new_cip_name = cip_change_message.original_cip.name
```
xswoodsx, line 8: should this be `cip_change_message.new_name.name`??

FriedSock: yes.. I originally had this bug but I fixed it when I changed these into helper methods below. So they don't need to be here any more 😅

app/controllers/cip_change_alert_controller.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class CipChangeAlertController < ApplicationController
+  before_action :authenticate_user!
+
+  def start
+  end
+
+  def acknowledge_start
+    current_user.cip_change_message&.destroy!
```
kicferk1, line 10: Do we need `current_user` here? `cip_change_message` is a helper method already :)

app/forms/cip_change_alert_form.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class CipChangeAlertForm
+  include ActiveModel::Model
+
+  attr_accessor :view_guidance_option
+
+  validates :view_guidance_option, presence: { message: "TODO fill me in, I am not real copy" }
```
kicferk1, line 8: I am not a real copy, but maybe the fact that this needs to be filled in can be stored in JIRA, and we can put a message here that looks reasonable to the user :D 

"Select whether you want to view guidance"?

app/views/cip_change_alert/guidance.html.erb
```diff
@@ -0,0 +1,32 @@
...
+      <li>
+        bullet item
+      </li>
+    </ul>
+
+    <%= form_for :cip_change_alert, url: cip_change_alert_guidance_path, id: "cip-change-alert-form", method: :post do |form| %>
+      <%= form.govuk_submit "Continue" %>
+    <% end %>
+  </div>
+</div>
```
kicferk1, line 31: I love that 😂 

Can we maybe not push that to prod until we get the right copy? :D

app/views/cip_change_alert/guidance.html.erb
```diff
@@ -0,0 +1,32 @@
...
+    <ul class="govuk-list govuk-list--bullet">
+      <li>
+        bullet item
+      </li>
+      <li>
+        bullet item
+      </li>
+    </ul>
+
+    <%= form_for :cip_change_alert, url: cip_change_alert_guidance_path, id: "cip-change-alert-form", method: :post do |form| %>
```
kicferk1, line 27: Let's use `form_with`, https://stackoverflow.com/questions/43868976/rails-5-form-for-vs-form-with

spec/cypress/integration/CipChangeAlert.feature
```diff
@@ -0,0 +1,29 @@
...
+    Given I am logged in as existing user with email "cip-change-early-career-teacher@example.com"
+    And I am on "dashboard" page
+    When I click on "link" containing "brief guide"
+    Then I should be on "cip change alert guidance" page
+    And I click on "continue button"
+    Then I should be on "training and support" page
+
+    When I navigate to "dashboard" page
+    And I click on "link" containing "brief guide"
+    Then I should be on "training and support" page
```
kicferk1, line 28: I think each page here should have 'and the page is accessible' and 'and percy is sent a snapshot called xxx' calls. 

Or at least each new page.

app/services/register_and_partner_api/sync_users.rb
```diff
@@ -105,6 +107,24 @@ def self.assign_user_attributes(attributes, user, profile)
       profile.cohort = get_cohort(attributes)
     end
 
+    def self.create_or_destroy_cip_change_message(user, attributes)
```
kicferk1, line 110: User CIPs can be nil - a quirk of how R&P enter data.

When we change from `nil` to something, we probably don't want to create a guidance message. When we change from something to `nil`, eh, I don't know, probably worth asking Isabelle.

app/views/cip_change_alert/guidance.html.erb
FriedSock: yeah, wasn't planning to merge this until the copy is sorted 😂

app/services/register_and_partner_api/sync_users.rb
FriedSock: Thanks, didn't know about this 👍

app/helpers/dashboard_helper.rb
```diff
@@ -8,4 +8,20 @@ def start_button_text(user)
...
     end
   end
+
+  def start_button_path(user)
+    if user.cip_change_message.present?
+      user.participant_profile.guidance_seen? ? cip_change_alert_start_path : cip_path(user.core_induction_programme)
+    else
+      user.participant_profile.guidance_seen? ? cip_path(user.core_induction_programme) : guidance_question_path
+    end
+  end
```
kicferk1, line 18: If user didn't see the guidance, we will never make the new `cip_change_message` for them, right?

So if the message is present, we know they saw the guidance, and therefore should go to `cip_change_alert_start_path` without checking anything.

Am I missing anything?

app/views/cip_change_alert/start.html.erb
```diff
@@ -0,0 +1,23 @@
...
+      Your school has changed your training and support programme from <%= original_cip_name %> to <%= new_cip_name %>
+    </p>
+
+    <p>
+      You can no longer access the materials you were previously using
+    </p>
+
+    <p>
+      Speak to your mentor for more information
+    </p>
```
kicferk1, line 16: Do we want dots at the end of sentences here?

app/views/cip_change_alert/start.html.erb
```diff
@@ -0,0 +1,23 @@
...
+
+    <%= form_with model: @cip_change_message, url: cip_change_alert_start_path, id: "cip-change-alert-form", method: :post do |form| %>
+      <%= form.govuk_submit "Continue" %>
+    <% end %>
+  </div>
+</div>
+
```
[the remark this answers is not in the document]
kicferk1: Ah, they point at a different action so we can redirect them to a different location after acknowledging the message. Alright, makes sense to have them both then.

app/views/cip_change_alert/start.html.erb
xswoodsx: I thought that but it's not on the design.

xswoodsx: tell a lie, yes it is. My bad

app/helpers/dashboard_helper.rb
xswoodsx: ok, yeah that's right, i hadn't read the user sync properly enough.

xswoodsx: i'll switch it back.

kicferk1 (approved):

PR #250, opened 2021-09-23
CPDEL-819 reporting to bigquery

FriedSock:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDEL-819

Just adds one github workflow for exporting data from staging (for testing). Will follow up with the others once this is working.

### Review Checks


.github/workflows/export_course_lesson_data_report.yml
```diff
@@ -0,0 +1,59 @@
...
+        shell: bash
+        run: |
+          cf install-plugin conduit -f
+
+      - name: Generate dashboard report
+        shell: bash
+        env:
+          PAAS_ORGANISATION: dfe
+          PAAS_SPACE: early-careers-framework-staging
+          ENV_STUB: staging
```
kicferk1, line 33: You can see how it works in a review app too, right?

kicferk1: Or maybe not, cause a workflow might need to be on develop to run it :P 
How about using `dev` then?

.github/workflows/export_course_lesson_data_report.yml
```diff
@@ -0,0 +1,59 @@
...
+        env:
+          CF_USERNAME: ${{ secrets.GOVPAAS_STAG_USERNAME }}
+          CF_PASSWORD: ${{ secrets.GOVPAAS_STAG_PASSWORD }}
+          PAAS_ORGANISATION: dfe
+          PAAS_SPACE: early-careers-framework-staging
+        run: |
+          cf api https://api.london.cloud.service.gov.uk
+          cf auth
+          cf target -o "${{ env.PAAS_ORGANISATION }}" -s "${{ env.PAAS_SPACE }}"
+          cf conduit ecf-engage-and-learn-staging-db -c '{"read_only": true}' -- psql -d rdsbroker_79c0fbce_1680_46f1_8ff0_5b176da3d6d1 -c "copy(select data from reports where identifier='course_lesson_data') to STDOUT" | sed 's/\\n/\n/g' > ~/data.csv
```
kicferk1, line 54: Is there some way we could use the same job to generate the data we store in our db as the data we send to big query? As in, why have a service generating the csv, if you are going to use conduit to do it in the real scenario?

FriedSock: yeah I was thinking about that, having some command that would just output a csv to stdout, but it would be difficult to prevent rails spewing crap out like warnings and deprecation messages

.github/workflows/export_course_lesson_data_report.yml
FriedSock: What is the difference in doing it with dev over staging if we have to merge the PR to test it?

kicferk1: Dev gets deployed automatically, staging you need to deploy yourself :D

.github/workflows/export_course_lesson_data_report.yml
FriedSock: I guess we could write to a tempfile and cat that instead

.github/workflows/export_course_lesson_data_report.yml
FriedSock: updated

kicferk1 (approved):
Let's see it in action :)

PR #263, opened 2021-12-01
Replace PERSONAL_TOKEN in deploy to prod workflow

FriedSock:
These steps are failing with 404 errors. Best guess is that this personal token belonged to @kicferk1 who has now been removed from the github org.
We should be able to use GITHUB_TOKEN instead https://docs.github.com/en/actions/security-guides/automatic-token-authentication, and this is
already used in an earlier step in the workflow.

## Ticket and context

Ticket:

### Review Checks

### DIFF
--- a/.github/workflows/deploy_to_production.yml
+++ b/.github/workflows/deploy_to_production.yml
@@ -62,7 +62,7 @@ jobs:
           restrictions: |
             null
         env:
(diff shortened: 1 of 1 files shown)


FriedSock:
This won't work because it isn't possible to give GITHUB_TOKEN enough permissions to remove branch protection. Also, there doesn't seem to be a reason why we need to update the master branch like this anyway.

